PR #58, opened 2026-01-14
chore: fix warnings

Sietse2202:
Fixed all warnings, except for one, that being a warning for `num-bigint-dig` using `vec!`

### DIFF
--- a/crates/backend/src/backend.rs
+++ b/crates/backend/src/backend.rs
@@ -1,28 +1,45 @@
 use std::{
-    collections::{HashMap, HashSet}, io::Cursor, path::{Path, PathBuf}, sync::Arc, time::{Duration, SystemTime}
+    collections::HashMap,
+    io::Cursor,
+    path::{Path, PathBuf},
+    sync::Arc,
+    time::{Duration, SystemTime},
 };
 
 use auth::{
     authenticator::{Authenticator, MsaAuthorizationError, XboxAuthenticateError},
-    credentials::{AccountCredentials, AUTH_STAGE_COUNT},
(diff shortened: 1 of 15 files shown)


enkvadrat:
Should probably have been split into 2 commits, one that formats the code and one that remove warnings, also aren't 99% of them only about unused imports? I think those are fine since this is in development right now and they will probably shift around.
What are the changes made here that is not covered by running `cargo fmt` committing the result and then `cargo fix`?

Sietse2202:
My editor auto formatted the places I fixed the warnings; I only wanted to
do that at first, and then do the formatting in a second commit

Moulberry:
Sorry, formatting-only changes aren't accepted. I can just run cargo fmt myself if needed

enkvadrat:
I was intending to do some contributions but since 
most editor (at least mine) format on save all my commits gets cluttered with formatting changes.

I could disable the format on save, but that will only make the formatting problem worse since now I need to do it by hand.
maybe you should run `cargo fmt` and then turn on format on save/commit.


```sh
➜  PandoraLauncher git:(master) ✗ cargo fmt
➜  PandoraLauncher git:(master) ✗ echo $(git diff --shortstat)
71 files changed, 3739 insertions(+), 2373 deletions(-)
```
If you use RustRover you should maybe turn on "Reorganize imports", "code cleanup"
and "Do not keep line breaks"
https://www.jetbrains.com/help/rust/rustfmt.html#format-file




> Sorry, formatting-only changes aren't accepted. I can just run cargo fmt myself if needed

PR #63, opened 2026-01-14
Add metainfo and desktop file

fluoriteByte:
says it in the title
[ do not merge yet until metafile is finalised, will say when it is ]


fluoriteByte:
@Moulberry I think im done in this side, now just need to figure out the flatpak cargo gen error
tell me if this looks good :3

scripts/helper_flatpak.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if [ -z "$1" ]; then
+    echo "Missing version argument"
+    exit 1
+fi
+
+version=${1#v}
+
+mkdir -p dist
+echo '#!/bin/sh' > dist/PandoraLauncher-Linux # dummy file, will build during flatpak
```
Moulberry: What does this do? I don't understand

fluoriteByte: basically just leveraging cargo packager for making a desktop file and nothing else x3
cus they want me to have the desktop file upstream + metainfo

Moulberry: Why is that? I don't think the information will be changing

fluoriteByte: tbh yeah, i guess we just include the desktop file directly in package lol
gonna copy the desktop file from the flatpak directly lol
but gonna do it in morning its 3 15 am x3

fluoriteByte: removed it in latest commit

fluoriteByte:
this should be good enough now

fluoriteByte:
also, sorry for asking, but @Moulberry, could you please update Cargo lock to point to the latest zed repo version? it's giving me dependency conflict issues when just replacing the commit with the latest one

Moulberry:
> also, sorry for asking, but @Moulberry, could you please update Cargo lock to point to the latest zed repo version? it's giving me dependency conflict issues when just replacing the commit with the latest one

It has to be updated at the same time as https://github.com/longbridge/gpui-component
Eventually I will pull both gpui and gpui-component into the PandoraLauncher repo and manage it there, which should make it a bit easier to work with

Issue #83, opened 2026-01-16
Option for Disabling the Game Output Log from Opening Automatically


J4PC:
When starting the launcher the log always opens in a new window no matter if you want it or not. An option for that would be really nice.

<img width="332" height="580" alt="Image" src="https://github.com/user-attachments/assets/6741b3d9-7f8d-4b5f-9358-f070bc78a064" />

Moulberry:
68d6dc9698311fc6768399ab0ec70bdc493979c9

Issue #80, opened 2026-01-16
[Feature request] UI - console output in logs tab


Sleepy-TF:
Would love it if the console could be seen from the logs tab in the instance menu, since after you close it on game launch you can't access it again as far as I can tell.

enkvadrat:
Maybe this should be configurable as an option, since in some cases you would not want the chat messages to be included.

PR #87, opened 2026-01-16
feat: game output stuff

vajzii:
### **Fixed:**

**Issue #83** 
- Pretty much just removing check to work like desired

**Issue #80** 
- This issue is not handled how user described, but instead with openable output window
- Output in window is not handled in the best way but its efficient and working one


vajzii:
### **Closed, opened new one with more stuff**

#90

Issue #82, opened 2026-01-16
No App Icon on Windows 11


J4PC:
I downloaded the currently newest version from github (PandoraLauncher-Windows_2.5.0_x64-setup.exe ) installed it and as can be seen in the attached image there is currently no icon. I can't be sure if this is a general issue or it just happened to me but in any case I wanted to mention it here seeing that an app without any icon looks really weird and ugly

<img width="455" height="731" alt="Image" src="https://github.com/user-attachments/assets/7665046f-5221-44e1-952a-e390d0bafa82" />

Moulberry:
5922b99f16c11ca5e876545a1fce839cea868153

PR #90, opened 2026-01-17
Game Output & Windows Icon

vajzii:
### **Fixed:**

**Issue #83** 
- Pretty much just removing check to work like desired

**Issue #82**
- App is now visible with pandora icon

**Issue #80** 
- This issue is not handled how user described, but instead with openable output window
- Output in window is not handled in the best way but its efficient and working one


Moulberry:
I'm not sure what you mean by
> Pretty much just removing check to work like desired

Firstly, what check? I assume you mean checking whether backend_config is set?
In what way was the check not working?

This PR contains a lot of confusing refactors, especially to log_reader.rs. Like why is raw_text being renamed to rawtext?

I'm not sure about keeping track of all of the lines when the log output window isn't open. This increases the memory footprint of the application.

vajzii:
> Firstly, what check? I assume you mean checking whether backend_config is set? In what way was the check not working?

While launching launcher outside of IDE, the option for opening game output was not available.

>This PR contains a lot of confusing refactors, especially to log_reader.rs. Like why is raw_text being renamed to rawtext?§

This could be simply explained and renamed back, it was little confusing for me since it was not my syntax style

> I'm not sure about keeping track of all of the lines when the log output window isn't open. This increases the memory footprint of the application.

I was thinking about setting max buffer size for this purpose

Moulberry:
> While launching launcher outside of IDE, the option for opening game output was not available.

This is because the commit that added the option has not been included in a release.

> This could be simply explained and renamed back, it was little confusing for me since it was not my syntax style

I would much prefer it if the changes are kept minimal

> I was thinking about setting max buffer size for this purpose

Perhaps. I'll have to think about this more.

Can you make a separate PR with only the windows manifest changes?

vajzii:
> Can you make a separate PR with only the windows manifest changes?

Yes, of course,

> I would much prefer it if the changes are kept minimal

Sure, will keep in mind.

vajzii:
> Can you make a separate PR with only the windows manifest changes?

Please check #91

Moulberry:
The other PR was merged

PR #120, opened 2026-01-27
add support for mangohud and gamemode on linux

FelipeFMA:
it only shows on linux and it's grayed out if the user doesn't have the package installed

image

image


Moulberry:
Thank you for your PR

PR #138, opened 2026-01-29
feat: add install script for linux

feraxhp:
A simple bash script to install the _"Portable executable"_ for linux. It downloads the executable, and creates de [Desktop Entry] file for the application.

the idea is to allows linux users install the launcher by just runing
~~~bash
curl -L "https://raw.githubusercontent.com/feraxtx/pandora/refs/heads/scripts/install/scripts/install.sh" | bash
~~~

> [!Note]
> The **url** is targeting the fork. 
> when if added to the repo, the url have to be change it for the _real repo_ url.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+
+load() {
+    while true; do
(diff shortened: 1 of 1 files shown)


Moulberry:
curl | bash, while maybe common for some developer tools, is a very dangerous practice. Users should not be made comfortable with running curl | bash, instead they should use their package manager. If there is no package, then a flatpak or appimage is still much better.

PR #148, opened 2026-01-30
Fix icon and hardcoded name in worldgen category button

Rajdacz:
- Moved hardcoded worldgen name to locales
- Fixed missing icon on the worldgen mod category button
Missing icon screenshot

### DIFF
--- a/crates/frontend/locales/locales.yml
+++ b/crates/frontend/locales/locales.yml
@@ -117,6 +117,8 @@ gui:
   en: GUI
 worldgen:
   en: World Generation
+worldgen-short:
+  en: Worldgen
 multiplayer:
   en: Multiplayer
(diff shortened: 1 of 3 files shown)


Moulberry:
The way the name is set up is intentional. "World Generation" is the correct name that Modrinth uses, and shows elsewhere (i.e. in the project tags). "Worldgen" is the shortened name used only for the filtering sidebar to save space.

The icon fix is fine, but the naming should stay.

Rajdacz:
Updated the PR:
- Added short translation to worldgen instead of replacing it
- Made a ts_short! macro to automatically handle shortened labels
- Fixed the icon for the worldgen category

Moulberry:
Thank you for your PR

PR #151, opened 2026-01-31
feat: implement auto-theme detection

### LINKED ISSUE #141: [Feat Request] Add auto theme
Allow __Pandora__ to change the dark-light theme with the system.

And add a _select light theme_ and _select dark theme_.

Fen1kks:
I apologize in advance if it seems a bit amateurish. I'm just starting out and wanted to contribute and improve myself. There are some minor errors, but there are no functional issues.
Closes #141

### DIFF
--- a/crates/frontend/src/interface_config.rs
+++ b/crates/frontend/src/interface_config.rs
@@ -22,7 +22,7 @@ impl gpui::Global for InterfaceConfigHolder {}
 pub struct InterfaceConfig {
 
     #[serde(default = "default_theme_selection", deserialize_with = "schema::try_deserialize")]
-    pub theme: SharedString,
+    pub active_theme: SharedString,
     #[serde(default, deserialize_with = "schema::try_deserialize")]
     pub main_page: SerializedPageType,
     #[serde(default, deserialize_with = "schema::try_deserialize")]
@@ -86,7 +86,7 @@ impl InterfaceConfigHolder {
     }
 }
 fn default_theme_selection() -> SharedString {
-    "System Default".into()
+    "Default Dark".into()
 }
 
 pub(crate) fn try_read_json<T: std::fmt::Debug + Default + for <'de> Deserialize<'de>>(path: &Path) -> T {
--- a/crates/frontend/src/modals/settings.rs
+++ b/crates/frontend/src/modals/settings.rs
@@ -21,8 +21,12 @@ pub fn build_settings_sheet(data: &DataEntities, window: &mut Window, cx: &mut A
     let settings = cx.new(|cx| {        
 
         let theme_select = cx.new(|cx| {
-            let mut themes: Vec<SharedString> = ThemeRegistry::global(cx).themes().keys().cloned().collect();
-            themes.insert(0, "System Default".into());
+            let mut themes: Vec<SharedString> = ThemeRegistry::global(cx)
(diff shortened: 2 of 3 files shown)


Moulberry:
You just removed custom themes entirely

Fen1kks:
Have I corrected it the way you wanted? @Moulberry

Moulberry:
Is the three separate dropdowns necessary? Couldn't it be combined into a single theme dropdown, with "Default System" just being one of the options?

Fen1kks:
I'll try.

Fen1kks:
If I understood you correctly, that's what you wanted. @Moulberry

crates/frontend/src/modals/settings.rs
```diff
@@ -18,28 +18,33 @@ struct Settings {
...
     let theme_folder = data.theme_folder.clone();
-    let settings = cx.new(|cx| {
-        let theme_select_delegate = SearchableVec::new(ThemeRegistry::global(cx).sorted_themes()
-            .iter().map(|cfg| cfg.name.clone()).collect::<Vec<_>>());
+    let settings = cx.new(|cx| {        
 
         let theme_select = cx.new(|cx| {
-            let mut state = SelectState::new(theme_select_delegate, Default::default(), window, cx).searchable(true);
-            state.set_selected_value(&cx.theme().theme_name().clone(), window, cx);
+            let mut themes: Vec<SharedString> = ThemeRegistry::global(cx).themes().keys().cloned().collect();
```
Moulberry: Should use `sorted_themes().iter().map(|cfg| cfg.name.clone())` instead of the raw keys

crates/frontend/src/interface_config.rs
```diff
@@ -15,10 +16,13 @@ struct InterfaceConfigHolder {
...
 
+
+
 #[derive(Debug, Default, Serialize, Deserialize)]
 pub struct InterfaceConfig {
-    #[serde(default, deserialize_with = "schema::try_deserialize")]
-    pub active_theme: SharedString,
+
+    #[serde(default = "default_theme_selection", deserialize_with = "schema::try_deserialize")]
+    pub theme: SharedString,
```
Moulberry: This shouldn't be renamed, because it will break people's existing configurations

crates/frontend/src/interface_config.rs
```diff
@@ -81,6 +85,9 @@ impl InterfaceConfigHolder {
         _ = write_safe(&self.path, &bytes);
     }
 }
+fn default_theme_selection() -> SharedString {
+    "System Default".into()
```
Moulberry: I don't think System Default should be the default, many people have a light system theme but still prefer their applications to be dark by default

crates/frontend/src/theme_utils.rs
```diff
@@ -0,0 +1,23 @@
+use crate::interface_config::{self, InterfaceConfig};
+use gpui::*;
+
+pub fn update_theme(cx: &mut App) {
+    let config = InterfaceConfig::get(cx);
+    let theme_name = if config.theme == "System Default" {
```
Moulberry: Try putting "System Default" into a const variable inside theme_utils, and then use that const variable inside settings.rs. Less error-prone

Moulberry:
Thank you for your PR

PR #163, opened 2026-02-01
fix flatpak metainfo

fluoriteByte:
fixed some missing stuff in the metainfo and invalid launchable id


fluoriteByte:
if you need me to change some screenshot captions, please tell me

fluoriteByte:
aaaa i keep forgetting to do skip ci x.x

fluoriteByte:
reasoning for that categories change: https://specifications.freedesktop.org/menu/latest/additional-category-registry.html

PR #167, opened 2026-02-03
Filter Modrinth search by instance version and allow vanilla loaders to install mods

castledking:
[Demonstration](https://youtu.be/j_ad30yUlLw)

- When adding to an instance, add facets so API returns only compatible mods/modpacks/shaders
- Use instance game version (versions:) and allow vanilla loaders to install mods
- No extra API calls; filtering is server-side via existing search
- Resourcepacks excluded from version filtering (unchanged behavior)

### DIFF
--- a/crates/frontend/src/modals/modrinth_install.rs
+++ b/crates/frontend/src/modals/modrinth_install.rs
@@ -161,28 +161,31 @@ fn open_from_entity(
 
                 let minecraft_version = instance.configuration.minecraft_version.as_str();
                 let instance_loader = instance.configuration.loader;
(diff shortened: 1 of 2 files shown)


Moulberry:
It should automatically switch the instance to fabric when you add a fabric mod. The loader_hint in the content install must not be getting set correctly. I'll look into this after

Moulberry:
Can you explain why version filtering applies to shaders in modrinth_page.rs, but shaders are listed as allow_all_versions in modrinth_install.rs?

I'm not sure which behaviour I prefer, but it should at least be consistent I think

PR #168, opened 2026-02-03
Modrinth Add from Modrinth: dynamic Installed/Install, update nub

castledking:
[Demonstration](https://youtu.be/8dp5bvLNh6w)

- Dynamic Install vs Installed state; update-check nub button; consistent button alignment
- Backend: push InstanceModsUpdated after InstallContent so UI updates without leaving view

### DIFF
--- a/crates/frontend/src/pages/modrinth_page.rs
+++ b/crates/frontend/src/pages/modrinth_page.rs
@@ -453,27 +453,15 @@ impl ModrinthSearchPage {
                             if project_type != ModrinthProjectType::Other {
                                 match &main_action {
                                     PrimaryAction::Install | PrimaryAction::Reinstall => {
-                                        if let Some(install_for_id) = install_for {
-                                            crate::modals::modrinth_install::open_version_picker(
-                                                name.as_str(),
-                                                project_id.clone(),
-                                                project_type,
-                                                install_for_id,
-                                                &data,
-                                                window,
-                                                cx,
-                                            );
-                                        } else {
-                                            crate::modals::modrinth_install::open(
-                                                name.as_str(),
-                                                project_id.clone(),
-                                                project_type,
-                                                None,
-                                                &data,
-                                                window,
-                                                cx,
-                                            );
-                                        }
+                                        crate::modals::modrinth_install::open(
+                                            name.as_str(),
+                                            project_id.clone(),
+                                            project_type,
+                                            install_for,
+                                            &data,
+                                            window,
+                                            cx,
+                                        );
                                     },
                                     PrimaryAction::InstallLatest => {
                                         crate::modals::modrinth_install_auto::open(


castledking:
Fixed, more consistent version picker: 
image

crates/backend/src/backend_handler.rs
```diff
@@ -295,8 +295,20 @@ impl BackendState {
                 self.download_all_metadata().await;
             },
             MessageToBackend::InstallContent { content, modal_action } => {
+                let target = content.target.clone();
                 self.install_content(content, modal_action.clone()).await;
                 modal_action.set_finished();
+                if let InstallTarget::Instance(id) = target {
```
Moulberry, line 301: This will reload all content unnecessarily.
The proper way to do it would be to do `self.instance_state.write().reload_immediately.insert((instance_id, content_folder));`.
This will make it so that when the filesystem detects the changes, it'll push those changes immediately to the frontend

crates/frontend/src/modals/modrinth_install.rs
```diff
@@ -290,6 +319,187 @@ fn open_error_dialog(title: SharedString, text: SharedString, window: &mut Windo
     });
 }
 
+/// Opens a version-picker dialog (dropdown-style). When user selects a version, opens the full install modal with that version pre-selected.
```
Moulberry, line 322: I don't understand the point of this dialog.
The next screen allows people to select the mod version they want.
Besides, most of the time, selecting an older mod version isn't what people want.

crates/frontend/src/pages/modrinth_page.rs
```diff
@@ -54,7 +55,7 @@ impl ModrinthSearchPage {
         if let Some(install_for) = install_for {
             if let Some(entry) = data.instances.read(cx).entries.get(&install_for) {
                 let instance = entry.read(cx);
-                can_install_latest = instance.configuration.loader != Loader::Vanilla;
+                can_install_latest = true;
```
Moulberry, line 58: The reason why this exists is because some mods provide multiple versions for different modloaders (e.g. forge and fabric).
By disallowing install latest, we force the person to select the modloader they want to use.
By removing this, the launcher will choose an arbitrary modloader which is confusing and bad behaviour

crates/frontend/src/pages/modrinth_page.rs
```diff
@@ -395,100 +427,159 @@ impl ModrinthSearchPage {
...
-                                            },
-                                        }
-                                    } else {
-                                        window.push_notification(
-                                            (
-                                                NotificationType::Error,
-                                                "Don't know how to handle this type of content",
-                                            ),
+                let primary_button_label = match &main_action {
+                    PrimaryAction::InstallLatest => "Install",
```
Moulberry, line 434: Why is this being hardcoded?

crates/frontend/src/pages/modrinth_page.rs
```diff
@@ -395,100 +427,159 @@ impl ModrinthSearchPage {
...
+                            cx.open_url(&format!(
+                                "https://modrinth.com/{}/{}",
+                                project_type, project_id
+                            ));
+                        }
+                    });
+
+                // Fixed left slot so both rows align; same fixed width for primary and Open Page so they match
+                let left_slot_w = px(40.0);
+                let button_w = px(118.0);
```
Moulberry, line 561: Hardcoding widths is not a good idea.
This will break when we add translations to the project, and is generally bad practice

crates/frontend/src/pages/modrinth_page.rs
```diff
@@ -542,55 +633,65 @@ impl ModrinthSearchPage {
...
         } else {
-            PrimaryAction::Install
+            (PrimaryAction::Install, None)
         }
     }
 }
 
-#[derive(PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
 enum PrimaryAction {
```
Moulberry, line 681: Most of these actions are now unused

Moulberry:
image
The "Downloads" text is now pretty badly misaligned (look at 93M Downloads)

Moulberry:
If the only point of the nub is to do updates, wouldn't it make more sense for it to be next to the "Install" button rather than the "Open Page" button? Overall I'm not totally convinced that the separate update button/nub is very elegant. Is there a specific problem that you're trying to solve?

PR #171, opened 2026-02-04
Fix instance rename bug

### LINKED ISSUE #169: [BUG] Cannot rename instances
Upon creating a instance, naming it works entirely fine, but going in to edit instance names results in:
"Unable to rename instance folder: Access is denied. (os error 5)",

Seemingly, this is due to insufficient file permissions. Disabling the anti-virus (that should be protecting file modification in certain directories) alongside adding the pandora directory to antivirus exclusions has no effect.

Image

Fen1kks:
I fixed the "Access Denied" error when renaming instances on Windows. The file watcher was locking the folder, so now it stops watching before rename and starts again after.

Tested and it works now.

Closes #169

### DIFF
--- a/crates/backend/src/backend.rs
+++ b/crates/backend/src/backend.rs
@@ -1025,11 +1025,56 @@ impl BackendState {
         }
 
         let new_instance_dir = self.directories.instances_dir.join(name);
+        let paths_to_unwatch = self.file_watching.read().get_instance_paths(id);
+        {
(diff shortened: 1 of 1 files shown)


Moulberry:
I appreciate you taking the time to submit PRs, but it seems like you are using AI as part of your workflow, which is creating an unnecessary burden on me to constantly review the mistakes that the AI creates

PR #175, opened 2026-02-04
Modrinth parameters panel tweaks

Rajdacz:
- Added missing resourcepacks categories icons
- Improved parameters panel width
- Added button for hiding categories
- Aligned categories buttons to the left

After
image

Before
image


Moulberry:
Why does it take up so much unnecessary space now?

Rajdacz:
In my opinion it looks clearer. On smaller window widths, it's smaller to not take up this space.

Moulberry:
> On smaller window widths, it's smaller to not take up this space.

What does this mean

Rajdacz:
It's responsive. Always uses 1/6 of the page (with min and max limit)

Moulberry:
I don't think that taking up the extra space is good.
The space is needed so you can read the descriptions on the right

Rajdacz:
I can change it if you want, but I made it to fit everything so I don't think this is problematic

Rajdacz:
Removed the commit with width changes
image

Moulberry:
Thank you for your PR

PR #181, opened 2026-02-10
Added default JVM Memory, JVM flags and jvm binary

### LINKED ISSUE #99: [Feature Request] Global setting overrides

SupiDupiToll:
I added an extra optiton in the syncing tab to configure default JVM Settings. When customsettings are set in an instance these are going to overwrite the settings in the syncing tab.

(and i needed to change the aux.rs to auxiliary.rs because the github action under windows would fail because it is a reserved windows file name)

would close #99 
image


Moulberry:
The syncing page is not the right place for this
Try to avoid formatting changes, it makes a lot of noise
JVM flags should be additive

PR #188, opened 2026-02-17
Extract hardcoded strings to locales

Rajdacz:
This PR extracts all hardcoded strings from the `frontend` crate to locales file for future translations

### DIFF
--- a/crates/frontend/src/lib.rs
+++ b/crates/frontend/src/lib.rs
@@ -34,17 +34,23 @@ pub mod ui;
 rust_i18n::i18n!("locales");
 
 macro_rules! ts {
-    ($($all:tt)*) => {
-        SharedString::new_static(ustr::ustr(&*rust_i18n::t!($($all)*)).as_str())
+    ($key:literal) => {
+        SharedString::new_static(
+            ustr::ustr(&*rust_i18n::t!($key)).as_str()
+        )
+    };
+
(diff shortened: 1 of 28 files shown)


crates/frontend/src/lib.rs
```diff
@@ -35,19 +35,19 @@ rust_i18n::i18n!("locales");
 
 macro_rules! ts {
     ($($all:tt)*) => {
-        SharedString::new_static(ustr::ustr(&*rust_i18n::t!($($all)*)).as_str())
```
Moulberry, line 38: Why was this change made?

Rajdacz: The current ustr implementation causes unbounded memory growth.
Using `SharedString::new_static` with ustr interns every translation result into a global, permanent cache. If a translation uses dynamic arguments, every unique string generated will stay in RAM forever

Moulberry: Then it should be changed so that only the dynamic argument translations don't have interning.
This should be possible to do by simply adding another branch to the macro with $all:expr instead of tt.
Interning should be kept on in the expr branch, but disabled in the tt branch

Rajdacz: Makes sense, I’ve updated the macro and amended the commit

Moulberry:
Thank you for your PR

PR #191, opened 2026-02-22
[Feature] Add Support for Waywall and similar programs

fooeyround:
Waywall is a Wayland compositor for Minecraft speedrunning which wraps the game to add support for embedding NinjaBrainBot for finding strongholds and resolution hotkeying. 

To support Waywall two features were added as instance configurations:
 - Add `Wrapper Command` (`mangohud`/`gamemoderun` wrap this when enabled)
 - Add `Disable GL Threaded Optimizations` (Linux Only: Fix for a crash with Nvidia GPUs on Wayland)
 
 
Those who want to use Waywall would do the following to an instance:
- `Add Wrapper Command`: `waywall wrap --`
- Set `Use System GLFW` to Waywall's patched GLFW
- Check `Disable GL Threaded Optimizations`

### DIFF
--- a/crates/backend/src/backend.rs
+++ b/crates/backend/src/backend.rs
@@ -1068,6 +1068,7 @@ impl BackendState {
             loader,
             preferred_loader_version: None,
             memory: None,
+            wrapper_command: None,
             jvm_flags: None,
             jvm_binary: None,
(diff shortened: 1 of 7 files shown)


Moulberry:
Thank you for your PR

PR #192, opened 2026-02-23
Addition of proxys for the launcher

bllryy:
- Added the ability to use a proxy while in the launcher
- Extended the settings sheet to make it wider
- I tested with the one proxy that I have which is http
- i also added changes to the gitignore (because I use RustRover)

proxy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -673,6 +673,7 @@ dependencies = [
  "toml 0.9.11+spec-1.1.0",
  "ustr",
  "uuid",
+ "walkdir",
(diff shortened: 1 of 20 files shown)


Moulberry:
Is there a reason why this PR includes a bunch of .idea files?

bllryy:
> Is there a reason why this PR includes a bunch of .idea files?

I use rustrover as my ide and Rustrover made the .idea by default

Moulberry:
Thank you for your PR

PR #198, opened 2026-02-24
Add Linux keyring fallback for KDE/KWallet

cranberry3148:
Use keyring-based fallback on Linux to read/write/delete credentials via org.freedesktop.secrets when the primary Secret Service path fails (seen on some KDE/KWallet setups), improving reliability and logging when fallback is used.

### DIFF
--- a/crates/auth/src/secret.rs
+++ b/crates/auth/src/secret.rs
@@ -65,6 +65,17 @@ mod inner {
             })
         }
 
+        fn attributes_for_uuid<'a>(uuid: &'a str) -> Vec<(&'a str, &'a str)> {
+            vec![("service", "pandora-launcher"), ("uuid", uuid)]
+        }
+
+        fn map_keyring_error(error: keyring::Error) -> SecretStorageError {
+            match error {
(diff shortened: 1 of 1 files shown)


Moulberry:
The PR mentions that it uses org.freedesktop.secrets through the OS keyring. As far as I understand, this is wrong. org.freedesktop.secrets is accessed through dbus, which what oo7 is trying to do. The PR description mentions both "org.freedesktop.secrets" and the "Secret Service path", but these are the same things.

By using the keyring crate with linux-native-sync-persistent, this PR adds support for the **in-kernel key management system**. The keyring crate does this by wrapping https://crates.io/crates/linux_keyutils.

In other words, the fallback added by this PR isn't actually using KWallet to store the credentials at all.

With that out of the way, I guess I have two questions:
Do we actually know why KDE/KWallet is failing to work through org.freedesktop.secrets?
Secondly, do we even want to have a fallback that uses the in-kernel key management instead of the recommended user-level secret service (i.e. KWallet)?

cranberry3148:
I just tested the latest commit and it seems like either the update of oo7 to 0.6 or the change in secret.rs fixed the problem for me.

PR #200, opened 2026-02-24
Fix panic when reading log files with invalid UTF8

max54nj:
Fixes a panic that occurred when reading log files containing invalid UTF8 characters or encountering I/O errors:

Backend panicked at crates\backend\src\backend_handler.rs:747:57
Cannot block the current thread from within a runtime.

It seems like this mostly happens with Forge versions below 1.20.6.


Moulberry:
Thank you for your PR

PR #228, opened 2026-03-02
Add profile page with interactive 3D skin viewer and Mojang skin upload

6space7:
- Software 3D rasterizer for Minecraft player skins with drag-to-rotate
- Skin history tracking with persistent storage
- Apply/revert skins via Mojang API
- Support for both Classic and Slim skin models

Screenshot 2026-03-03 at 5 47 53 AM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6326,6 +6326,7 @@ dependencies = [
  "js-sys",
  "log",
  "mime",
+ "mime_guess",
  "native-tls",
  "percent-encoding",
  "pin-project-lite",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ paste = "1.0.15"
 quick-xml = "0.38.3"
 rand = "0.8.5"
 regex = "1.12.2"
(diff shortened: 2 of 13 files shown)


Moulberry:
Hm, I like the skin renderer, but I'm not sure about the rest of the changes.
I think a "Skin Library" is better than a "Skin History"
Importing from a username seems like a weird choice when importing from a file/url is missing
I can see that AI was used, which is probably contributing to some of the weird design decisions.

Maybe you could make a PR with only the software skin renderer, and I can do the rest of the design for the page/controls?

6space7:
> Hm, I like the skin renderer, but I'm not sure about the rest of the changes. I think a "Skin Library" is better than a "Skin History" Importing from a username seems like a weird choice when importing from a file/url is missing I can see that AI was used, which is probably contributing to some of the weird design decisions.
> 
> Maybe you could make a PR with only the software skin renderer, and I can do the rest of the design for the page/controls?

Sounds good. And yeah, I used AI for the UI since design isn’t really my strong side. I usually rely on AI for that part. I’ll open another PR that only includes the skin renderer

Moulberry:
Closed as discussed

Issue #214, opened 2026-02-28
Register imported mods in content library


boomshroom:
I was hoping to be able to deduplicate the mods of my existing PrismLauncher instances, but it seems that they only deduplicate if there's already a mod in the content library with a matching hash, rather than registering the new mods to be used later in the import process.

Moulberry:
Afaik it doesn't deduplicate the mods even if you have it in your content library.
The importer just copies over the `.minecraft` (or `minecraft`) folder one-to-one.
I'm not 100% sure if it's a good idea to apply the deduplication to the import

boomshroom:
Another option would be to have a button to substitute a given local mod with a Modrinth equivalent. It'd still be manual, but it wouldn't require removing each mod and then finding it again to reinstall it.

PR #232, opened 2026-03-04
Mini import refactor and support for AtLauncher

dragmine149:
Add support for importing instance and accounts from `AtLauncher` (not flatpak version)

Refactor of the launcher imports was only done to reduce amount of complexity when adding more launchers in the future. Nothing too complex just seperated code into their own files and made the front-end launcher page loop through the potential launchers instead of hardcoded them one by one. I have not tested the other 3 launchers but as i didn't change anything logic wise i don't see why they should break.

The main feature here is support for AtLauncher. Which i fell like there are still a few things i've missed but it works enough for this PR.

Instance Importing
- import instance
- import icon
- import *most* settings (some settings i couldn't work out a direct translation for)
 *The settings not translated are game flags/jvm params and paths of custom glfw/openal. There were also some extra settings atlauncher has like `permgen` and `pre-launch/post-exit commands` which weren't translated due to not existing here*

Account Importing
- import account
- import auth token
- select last account
*Note: for some reason you have to re-sign in anyway, idk if this is intentional or if i've just failed to translate something that is required.*

I've left a lot of commented out code for the structs i use to parse the json, some of which is curseforge, modrithn and atlauncher respectiable. Even if we don't fully use these now, it might be useful to keep them around for a future use case, alternatively i could just remove the unused attributes all together.

The `.zed` folder is also include as per your dislike of `cargo fmt` i've seen on other PRs. (This was the easiest way for me not to continue formating the code every time i opened a file or smth)

And finally, my thoughts on your comment here: https://github.com/Moulberry/PandoraLauncher/issues/214#issuecomment-3976678347
>  I'm not 100% sure if it's a good idea to apply the deduplication to the import

Personally, i would say deduplicate mods which we know the source from. For example, AtLauncher provides a big list of mods in the instance file and each mod contains files such as `modrinthProject` or `curseForgeProject` which would be perfect for verifying the source of said mods.


crates/backend/src/launcher_import/atlauncher.rs
```diff
@@ -0,0 +1,483 @@
...
+// #[serde(rename_all = "case")]
+struct AtLauncherDisplayClaim {
+	uhs: String,
+}
+
+
+
+
+pub async fn import_from_atlauncher(backend: &BackendState, path: &Path, import_accounts: bool, import_instance: bool, modal_action: ModalAction) {
+	// probably a better way of doing this mess...
```
Moulberry, line 251: The better way is:
```rust
let Ok(launcher_config_bytes) = std::fs::read(path.join("configs/ATLauncher.json")) else {
    return;
}
```

Moulberry:
Thank you for your PR

crates/backend/src/launcher_import/atlauncher.rs
dragmine149: tbh, this was probably an artifact left behind as sometimes i would make `import_from_atlauncher` return something other than `()` and sometimes i needed to be able to panic so i could see why it failed.

As you've already merged i won't change it now but will keep in mind for future prs

Issue #233, opened 2026-03-04
[Feature Request] Show MOTD and oline player count on quick play page


Adrien5902:
Modrinth launcher displays servers motd as well as online player count in quick play page
<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/7899c87d-1a34-4ea5-a55f-70d228d06d97" />
which i'd like to see too in Pandora
<img width="1538" height="896" alt="Image" src="https://github.com/user-attachments/assets/9209cebd-5ef8-4deb-b5b8-242a23536c2f" />

Moulberry:
22e51fe16b6bfc855b6433061e4da6425f83715c

PR #194, opened 2026-02-23
Closes #177, Adds color to the resource pack descriptions

Revilo0509:
This pr fixes issue #177.
It also includes
* Adding .vscode to gitignore
* Adding a try-statement around the scanner in the launch wrapper.
* Change function "create_description" to use arguments by reference instead of ownership

The biggest changes are mostly formatting by running cargo fmt

Issue #177, opened 2026-02-06
Properly Render Minecraft Color Codes in Resource Packs


harmfulladvocado:
When using resource packs with custom menus or UI, color codes like §6, §5, etc., are displayed as raw text instead of colored formatting.

Feature Request:
- Detect and render Minecraft color codes in menu/UI text, so they appear colored as intended (e.g., §6 = gold, §5 = purple)
- Prevent raw codes from appearing to users in any launcher menus/resource pack previews

This would provide a cleaner look in custom menus and make resource packs easier to use and read.

Example:
- Instead of showing text like "§6News" in the menu, show "News" in gold color.

<img width="462" height="259" alt="Image" src="https://github.com/user-attachments/assets/d45a1a24-a35d-4865-ae33-01e78e3b25c0" />

PR #237, opened 2026-03-06
feat: server pings for MOTD and player count

Maganoos:
vibecoded slop that sort of works but lowkey breaks formatting to implement #233. (don't get mad C:)
i'm not sure how to resolve the formatting issue other than decreasing font size or line width
idk man :smile:

### DIFF
--- a/crates/backend/src/instance.rs
+++ b/crates/backend/src/instance.rs
@@ -953,7 +953,6 @@ async fn load_servers_summary(server_dat_path: &Path) -> anyhow::Result<Vec<Inst
         name: Arc<str>,
         ip: Arc<str>,
         png_icon: Option<Arc<[u8]>>,
-        addr: String,
     }
 
     let mut entries = Vec::with_capacity(servers.len());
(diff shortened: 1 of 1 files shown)


crates/backend/src/instance.rs
```diff
@@ -970,18 +978,215 @@ fn load_servers_summary(server_dat_path: &Path) -> anyhow::Result<Vec<InstanceSe
...
+                        out.push_str(&extract_chat_text(child));
+                    }
+                }
+                out
+            }
+            _ => String::new(),
+        }
+    }
+
+    fn strip_section_codes(s: &str) -> String {
```
Adrien5902: I feel like this shouldn't have been implemented inside this function here because we might end up reusing it when implementing resource pack description coloring see #177

Maganoos: probably yeah, just wanted something quick and easy, not sure how I would do that 😁

crates/backend/src/instance.rs
```diff
@@ -970,18 +978,215 @@ fn load_servers_summary(server_dat_path: &Path) -> anyhow::Result<Vec<InstanceSe
...
+                    .decode(v)
+                    .map(Arc::from)
+                    .ok()
+            });
+
+        let addr = if ip.contains(':') {
+            ip.to_string()
+        } else {
+            format!("{ip}:25565")
+        };
```
Adrien5902: I don't understand why you're adding port here to then strip later in line 1083
I would remove addr from server entry then pass ip to async_ping_server and format addr in async_ping_server like so

```rust
    let (host, port): (&str, u16) = if let Some(pos) = ip.rfind(':') {
        let h = &ip[..pos];
        let p = ip[pos + 1..].parse::<u16>().unwrap_or(25565);
        (h, p)
    } else {
        (ip, 25565)
    };

    let addr = format!("{host}:{port}");
```

also you should make a const for 25565 since it's used twice here

Adrien5902:
UI isn't that great too, here hypixel ip address gets messed up with the other server
image

crates/backend/src/instance.rs
Adrien5902: You could have left it with the § for now, just like in the resource pack ui, since coloring is being worked on in #194

Maganoos:
yeah, I'm aware, I think I stated in the original comment about that as well

Moulberry:
The architecture is completely wrong. Pinging shouldn't be done each time the servers.dat is loaded.
This needs proper component handling, not just grabbing 'text' from the root of the component.
This is probably slightly too much for an LLM to do

Maganoos:
alright :) just wanted to give it a shot

PR #245, opened 2026-03-07
Add "Hide Server Addresses" Privacy Setting

fooeyround:
This PR adds a privacy section to the interface config with a "hide server addresses" config.

This is helpful for those streaming or sharing screen while showing the launcher. 

Image:
image

Please let me know if you would like the setting inverted, somewhere else, or not present entirely.


Moulberry:
Thank you for your PR

Issue #243, opened 2026-03-07
GUI Overflow broken


sleepy-evelyn:
The text of a couple elements can overflow the box they are contained within. I also suspect that if a person has a long modpack name the `start` and `open .minecraft folder` may become invisible if a person doesn't have a wide enough screen. Ideally these buttons are placed below the modpack instances name or the modpack instance name box is given a maximum possible width.

<img width="318" height="194" alt="Image" src="https://github.com/user-attachments/assets/8f207dfe-9275-4201-8673-4532e8e1c666" />

<img width="144" height="133" alt="Image" src="https://github.com/user-attachments/assets/be9d703e-19fd-4658-a148-24491284ca30" />

<img width="872" height="68" alt="Image" src="https://github.com/user-attachments/assets/52b3e978-077b-4beb-a1c0-75f6b1b2ad50" />

dragmine149:
Personally, fells like this should be related to #240

Moulberry:
I mean there isn't really a good solution. If you want to make the window super small then yeah in some cases there won't be enough space to fit the text on your screen

perman24681012-dev:
maybe when text cant be visible then do "..." or just make that max smallest size of window bigger

EvieTheOwl:
> I mean there isn't really a good solution. If you want to make the window super small then yeah in some cases there won't be enough space to fit the text on your screen

This isn't the same issue. @dragmine149 linked it to #240 but it isn't related to that issue.
Text overflows containers if you reduce the horizontal width only by a small amount.

The point is the GUI is broken because text should never overflow a container box. There are a tonne of ways this can be prevented, scrolling text, `...` at the end of the word. These are very basic GUI design principles.

Simularly you shouldn't put buttons in a space dependent on the length of text used to describe a modpack. If those buttons gets pushed off the end of the screen thats bad UI design.

The `Sleepy_Evelyn` username also should not wrap. I did not make the horizontal width as small as the one shown in #240 and yet it still overflows.

dragmine149:
> This isn't the same issue. @dragmine149 linked it to https://github.com/Moulberry/PandoraLauncher/issues/240 but it isn't related to that issue.

I only linked it to this issue because of the horizontal width. Even if it's not the smallest size possible they are still related and could be fixed in one patch. It's still related though as both cases are to do with the `Button`  struct.

[...]

PR #248, opened 2026-03-08
fix: Delete button not showing on the account list for long names

dragmine149:
I noticed this whilst digging into #243 and hence fixed it

Screenshot_20260308_202003

*Left = before, right = after*

The main priority was getting the delete button to show, but got it wrapping anyway. 

Note: This will only happen for those accounts with weird usernames like the one shown.

### DIFF
--- a/crates/frontend/src/ui.rs
+++ b/crates/frontend/src/ui.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use bridge::{instance::InstanceID, message::MessageToBackend};
 use gpui::{prelude::*, *};
 use gpui_component::{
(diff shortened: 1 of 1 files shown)


Moulberry:
This should be fixed without the head icon going outside the button
With this change there is also wasted empty space to the right of the delete button now

dragmine149:
*after messing around and somehow getting something to work...*
Update the PR now:
- Delete buttons are aligned and fully to the right
- Heads are alligned, not fully to the left but good enough.
- Text is also alligned.

I did make the size of the buttons slightly bigger to compensate for the text wrapping overflowing. `size_12` instead of `size_10`

Screenshot_20260310_001436

Moulberry:
Fixed it in a cleaner way in 696ed9c90d8435c7be3b2bf2f18c413c5eaafeee

Issue #205, opened 2026-02-25
[Feature Request] Account Per Instance or Link Account to Instance


9gvuwu:
Add the ability to bind a specific Minecraft account to each individual instance, allowing automatic account switching when launching that instance.
Currently, all instances share the same active account. If a user operates multiple instances for different purposes (for example, one for regular gameplay and another for server administration with different mods), they must manually switch accounts each time.
This leads to:
Repetitive and unnecessary manual steps
Frequent mistakes (launching the wrong account in the wrong instance)
Having to fully restart Minecraft to change accounts
Wasted time and disrupted workflow
For users who manage servers or test modded environments, this becomes especially inconvenient.

Moulberry:
3acd504c85044a8c365d91be60b3d2e35058e576

PR #249, opened 2026-03-09
Link account to instance

dragmine149:
Implements: #205 (i needed a break from working on some importer logic for another pr so did this instead...)
Screenshot_20260309_033206

Adds a new option with the ability to choose a specific account saved to that instance.
Screenshot_20260309_033437

Accounts are read from the list of accounts, and is reset by "clean"-ing (that is what the code says...) the input box. Personally, i'm not a big fan with how you clear the account but i choose it due to being an official API provided by the Select interface. I'm happy to make changes to this in order to make it more obvious if need be.

Launching the game in any format will obviously use the chosen account. The downside is that we have to load the config file before we sign in, but i didn't fell anything different so i'm going to assume this is ok.


crates/bridge/Cargo.toml
```diff
@@ -18,3 +18,6 @@ relative-path.workspace = true
 sanitize-filename.workspace = true
 rustc-hash.workspace = true
 indexmap.workspace = true
+
+gpui.workspace = true
```
Moulberry: Do not depend on UI inside the bridge crate.
The bridge crate must be frontend/backend agnostic.

crates/frontend/src/pages/instance/settings_subpage.rs
```diff
@@ -373,6 +398,22 @@ impl InstanceSettingsSubpage {
...
+       	_cx: &mut Context<Self>,
+    ) {
+	   	let SelectEvent::Confirm(value) = event;
+
+		self.backend_handle.send(MessageToBackend::SetInstancePreferredAccount {
+			id: self.instance_id,
+			account: *value,
+		});
+
+   		println!("Selected: {:?}", value);
```
Moulberry: leftover debugging

crates/schema/src/instance.rs
```diff
@@ -12,6 +13,8 @@ pub struct InstanceConfiguration {
     pub loader: Loader,
     #[serde(default, skip_serializing_if = "crate::skip_if_none")]
     pub preferred_loader_version: Option<Ustr>,
+    #[serde(default, skip_serializing_if = "crate::skip_if_none")]
```
Moulberry: should probably have `deserialize_with = "crate::try_deserialize"`

crates/frontend/src/pages/instance/settings_subpage.rs
```diff
@@ -105,6 +107,19 @@ impl InstanceSettingsSubpage {
         }).detach();
         cx.subscribe(&version_select_state, Self::on_minecraft_version_selected).detach();
 
+        let account_select_state = cx.new(|cx| {
+        	let accounts = data.accounts.read(cx);
+         	let list = AccountList::from(accounts);
```
Moulberry: Use NamedDropdown here, should save a bit of boilerplate.
Just make a NamedDropdownItem with item being the uuid

dragmine149:
Updated everything requested to be updated.

PR #257, opened 2026-03-14
Importer improvements

### LINKED ISSUE #202: Support use of portable versions of MultiMC/Prism/etc
Portable versions of these programs exist, and are supported by the original maintainers, so they should also be supported by the import function.

dragmine149:
# Features
- Load from custom directory (should close #202)
- Select which instances you want to import instead of all of them
	*This also forces a refresh upon importing to update the UI*
- Show an error message if an instance can't be imported due to the name already existing.
	*Should we expand this to have an option to rename it in the importer instead of later?*
- Import button has a hover effect for additional information
Screenshot_20260314_143408

If requested i can try and split out the features, i kinda made some of the features whilst i was mentally working on how to make the custom import clean. 

# Bugs
Screenshot_20260314_144623

I tried to figure out how to fix this, but i couldn't work out an easy way as to how. I believe this would happen even without this update, this update just made it more prominent. 

# 
This PR is also one to request code review as i suspect there are definitally parts of my code that i can improve.

### DIFF
--- a/crates/backend/src/launcher_import/atlauncher.rs
+++ b/crates/backend/src/launcher_import/atlauncher.rs
@@ -412,7 +412,7 @@ fn import_instances_from_atlauncher(backend: &BackendState, instances: HashMap<P
 
     for (entry, state) in instances {
 	    if state != ImportStatus::Importing {
-			// println!("Not importing: {:?}", entry);
+            // println!("Not importing: {:?}", entry);
 		    continue;
 	    }
 
--- a/crates/backend/src/launcher_import/mod.rs
+++ b/crates/backend/src/launcher_import/mod.rs
@@ -7,8 +7,8 @@ use strum::IntoEnumIterator;
 use crate::{BackendState,
     launcher_import::{
         modrinth::{import_instances_from_modrinth, read_profiles_from_modrinth_db},
-	    multimc::{import_from_multimc, try_load_from_multimc},
-	    atlauncher::import_from_atlauncher
+        multimc::{import_from_multimc, try_load_from_multimc},
+        atlauncher::import_from_atlauncher
     }
 };
 
@@ -56,14 +56,14 @@ pub fn discover_instances_from_other_launchers(backend: &BackendState) -> Import
 
 /// Loop through all potential instances and returns the first one found.
 pub fn discover_instances_from_path(backend: &BackendState, path: PathBuf) -> Option<ImportFromOtherLauncher> {
- 	debug!("Received request to update data w/path: {:?}", path);
+    debug!("Received request to update data w/path: {:?}", path);
 
     // modrith doesn't conform to standards, hence we deal with it separately...
     if let Ok(modrinth) = read_profiles_from_modrinth_db(&path, &backend.directories.instances_dir) {
         if modrinth.is_some() { return modrinth; }
     }
 
-  	for launcher in OtherLauncher::iter()
+    for launcher in OtherLauncher::iter()
         .filter(|launcher| *launcher != OtherLauncher::Modrinth)
     {
         let details = get_launcher_details(backend, launcher, &path);
@@ -173,7 +173,7 @@ pub async fn import_from_other_launcher(backend: &BackendState, details: ImportF
             import_from_multimc(backend, details, modal_action).await;
         },
          OtherLauncher::ATLauncher => {
-          	import_from_atlauncher(backend, details, modal_action).await;
+            import_from_atlauncher(backend, details, modal_action).await;
         }
     }
 }
--- a/crates/backend/src/launcher_import/modrinth.rs
+++ b/crates/backend/src/launcher_import/modrinth.rs
@@ -149,7 +149,7 @@ pub fn read_profiles_from_modrinth_db(data_dir: &Path, pandora_dir: &Path) -> ru
         let path: String = row.get(0)?;
         let profile = profiles.join(path);
         if profile.is_dir() {
-        	let state = if pandora_dir.join(profile.file_name().unwrap()).exists() { ImportStatus::Duplicate } else { ImportStatus::Importing };
+            let state = if pandora_dir.join(profile.file_name().unwrap()).exists() { ImportStatus::Duplicate } else { ImportStatus::Importing };
             paths.insert(profile, state);
         }
     }
--- a/crates/backend/src/launcher_import/multimc.rs
+++ b/crates/backend/src/launcher_import/multimc.rs
@@ -397,13 +397,13 @@ fn import_instances_from_multimc(backend: &BackendState, instances: HashMap<Path
             continue;
         }
 
-    	if !entry.exists() {
-	        continue;
-	    };
-	    let folder = entry;
-	    if !folder.is_dir() {
-	        continue;
-	    }
+        if !entry.exists() {
+            continue;
+        };
+        let folder = entry;
+        if !folder.is_dir() {
+            continue;
+        }
 
         let Some(filename) = folder.file_name() else {
             continue;
--- a/crates/bridge/src/import.rs
+++ b/crates/bridge/src/import.rs
@@ -9,13 +9,13 @@ pub struct ImportFromOtherLaunchers {
 #[derive(Debug, Clone)]
 pub struct ImportFromOtherLauncher {
     // launcher is duplicated here for when we send a custom path request.
-	pub launcher: OtherLauncher,
-	pub instances: HashMap<PathBuf, ImportStatus>,
-	pub account: Option<PathBuf>,
-	pub custom_import: bool,
-	// This is placeholder for a future update if we ever implement it.
-	// might remove it before releasing this PR though... (if i don't, then i forgot.)
-	// pub can_deduplicate: bool,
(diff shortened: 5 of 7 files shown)


crates/backend/src/launcher_import/atlauncher.rs
```diff
@@ -244,31 +244,52 @@ struct AtLauncherDisplayClaim {
...
+pub fn is_valid_ataccount(path: &Path) -> Option<PathBuf> {
+    let account_path = path.join("configs/accounts.json");
+    account_path.exists().then(|| account_path)
+}
 
 
+pub async fn import_from_atlauncher(backend: &BackendState, details: ImportFromOtherLauncher, modal_action: ModalAction) {
+    let path =
+    if let Some(account_path) = &details.account {
+        Some(account_path.parent().unwrap().parent().unwrap())
```
Moulberry, line 261: This seems completely backwards, the root path of the instance should be what's stored, and the paths for the account/instances should be created from the root path.

crates/backend/src/launcher_import/atlauncher.rs
```diff
@@ -244,31 +244,52 @@ struct AtLauncherDisplayClaim {
...
+ 	let launcher_config = {
+        if let Some(config_path) = path {
+      		match std::fs::read(config_path.join("configs/ATLauncher.json")).ok() {
+      		    Some(launcher_config_bytes) => serde_json::from_slice::<AtLauncherConfig>(&launcher_config_bytes).expect("Failed to parse to json"),
+      		    None => {
+                    warn!("Failed to read config path for ATLauncher.json!");
+                    return;
+                },
+      		}
+        } else { AtLauncherConfig::default() }
```
Moulberry, line 276: Why do we even allow importing if the config doesn't exist?
The frontend should show an error with 'invalid instance' if the config doesn't exist

crates/backend/src/launcher_import/atlauncher.rs
```diff
@@ -244,31 +244,52 @@ struct AtLauncherDisplayClaim {
...
+	}
+	if !details.instances.is_empty() {
+		import_instances_from_atlauncher(backend, details.instances, &launcher_config, &modal_action);
+	}
 }
 
-async fn import_accounts_from_atlauncher(backend: &BackendState, path: &Path, launcher_config: &AtLauncherConfig, modal_action: &ModalAction) {
-    let tracker = ProgressTracker::new("Reading accounts.json".into(), backend.send.clone());
+async fn import_accounts_from_atlauncher(backend: &BackendState, account_path: &Path, launcher_config: &AtLauncherConfig, modal_action: &ModalAction) {
+	let tracker = ProgressTracker::new("Reading accounts.json".into(), backend.send.clone());
```
Moulberry, line 288: Please ensure indentation is correct. 4 spaces should be used, not tabs

crates/backend/src/launcher_import/atlauncher.rs
```diff
@@ -383,27 +404,25 @@ fn try_load_from_atlauncher(config_path: &Path, launcher_config: &AtLauncherConf
...
-    for entry in read_dir {
-        let Ok(entry) = entry else {
-            continue;
-        };
-        let folder = entry.path();
-        if !folder.is_dir() {
-            continue;
-        }
+    for (entry, state) in instances {
+	    if state != ImportStatus::Importing {
```
Moulberry, line 414: What is with this 'import status' system? It seems pointless, we should already know an import is in progress because of the modal

crates/backend/src/launcher_import/mod.rs
```diff
@@ -12,42 +15,127 @@ use crate::{BackendState, launcher_import::{
...
-    imports.imports[OtherLauncher::ATLauncher] = from_subfolders(&atlauncher_instances, &|path| {
-        path.join("instance.json").exists()
+    let atlauncher_instances = data_dir.join("atlauncher");
+    imports.imports[OtherLauncher::ATLauncher] = get_launcher_details(backend, OtherLauncher::ATLauncher, &atlauncher_instances);
+
+  imports
+}
+
+/// Loop through all potential instances and returns the first one found.
+pub fn discover_instances_from_path(backend: &BackendState, path: PathBuf) -> Option<ImportFromOtherLauncher> {
```
Moulberry, line 58: Why are we doing this? If the user wants to import from a custom directory, they should be selecting both the launcher and path

crates/backend/src/launcher_import/mod.rs
```diff
@@ -12,42 +15,127 @@ use crate::{BackendState, launcher_import::{
...
+fn get_launcher_details(backend: &BackendState, launcher: OtherLauncher, path: &PathBuf) -> Option<ImportFromOtherLauncher> {
+    let mut import_data = ImportFromOtherLauncher::new_launcher(launcher);
+
+    // we take the best case scenario and check the current folder.
+    let is_instance = instance_check(launcher, &path);
+    if is_instance {
+        import_data.instances.insert(path.clone(), ImportStatus::Importing);
+    } else {
+        // otherwise we check one level deep
+        for path in loop_subfolders(&path, &|path| instance_check(launcher, &path.to_path_buf())) {
```
Moulberry, line 106: Why are we doing all this stuff with subfolders?
This seems wasteful and extremely prone to breaking

crates/backend/src/launcher_import/mod.rs
```diff
@@ -61,49 +149,41 @@ fn from_subfolders(folder: &Path, check: &dyn Fn(&Path) -> bool) -> Option<Impor
...
+            import_from_multimc(backend, details, modal_action).await;
         },
         OtherLauncher::Modrinth => {
-            if import_instances {
-                let modrinth = data_dir.join("ModrinthApp");
-                if let Err(err) = import_instances_from_modrinth(backend, &modrinth, &modal_action) {
+            if details.instances.iter().any(|(_, status)| *status == ImportStatus::Importing) {
+                // In theory, (due to limitations) the parent, parent path of any instance should be our folder..
+                // if it's not, something went wrong. but at that point something else is probably completely broken...
+                let modrinth = (&details.instances).iter().nth(0).unwrap().0.parent().unwrap().parent().unwrap();
```
Moulberry, line 165: Same as the first comment, this is insane

Moulberry:
I stopped reviewing it at this point because the entire architecture of the system seems to have way too many issues

dragmine149:
To explain all the reviews at once... (as i should have done in my original post..)

## `ImportStatus`
This is:
- A bad naming system from me
- The important system behind showing the user if an instance will not be imported.
- The user telling the system which ones to import or not (although this could just be a boolean, the enum was to allow for the duplicated front-end feedback)

This feature is more of a QOL one, instead of just skipping over we can warn the user in advance. additional future updates could bring an option to rename before importing.
Alternatively, this feature could just be removed. Then i won't send the data back to the backend as to which paths to ultimately import.

## Indentation
*i thought i fixed this....* oh well, i'll make sure to double check next time.

## Parent folder/sub-folder mess.
I wanted to give the user more leniency as to which folder they chose, hence this mess...


Some cases i would argue have a legitament use. When they recieve data from the front-end as to which instances to import, the system doesn't know the parent folder if it needs one. As i'm writting this, i have realised i could just, send some more data as to the root path of the folder the user has picked back and forth as well... Hence this is not an issue.

#
Tomorrow (monday), i'll start working on trying to improve some of these features, i want to try and avoid ending up in a loop of, architecture issues hence its probably easier to restart some of these from scratch.
 
 I'll aim for the following things at least:
- reduce amount of `.parent()`. Probably by rewritting this system. 
- I've worked out recently, there are some segments of the code i can optimise, although probably not reviewed, there is definitally an improvement that can be done after the import request has been sent.
- Hopefully try and make a better architecture. Although i'm not fully sure yet on what this would entitle. Personally, i would have tried to make a `trait` for the individual importers to require, but between `async` and rust not doing exactly what i wanted, i kinda gave up with that idea.

Before i start working on improvements improvements, i Should ask.
Should i split it up this time instead of combining them all into one, or should i just continue with this PR and fix them one-by-one?
Additionally, based on the above, is there anything else to know/to whatever (one of those many words)

Moulberry:
Yes please try to split the changes. Perhaps just make 1 change at a time, then it can be reviewed etc, then the next change

dragmine149:
I've converted this PR to a draft and will leave it as such.
All the features i wish to implement i will open new PRs for and only close this original PR once they have been all been resolved.

ruri4:
if my PR (https://github.com/Moulberry/PandoraLauncher/pull/278) gets approved, could you also support importing of instance stats? so we can satisfy https://github.com/Moulberry/PandoraLauncher/issues/272

dragmine149:
> if my PR (#278) gets approved, could you also support importing of instance stats? so we can satisfy #272

There are numerous other importer related things that need to be done. Even more than the list in this PR. So as long as it's close to a 1-to-1 sync per stat, it shouldn't be that hard to do.

dragmine149:
Seems to have been made mostly redundant with cd65af4, hence i'll close this.

The only things missing from that commit are QoL Features like `un/check all` and #286

Issue #272, opened 2026-03-15
[Feature Request] Playtime tracking


CygnusGD:
Prism Launcher has a feature where it tracks playtime per-instance and can show all of them summed up as total playtime, which I'd like to see in Pandora Launcher. Copying over playtime when importing instances from Prism Launcher (or other launchers that track playtime) would also be really nice.

bllryy:
I might try and see if I could implement this.

PR #278, opened 2026-03-16
feat: Session statistics

notruri:
## Description

adds stats into instances that can track playtime, session count, and last played.

session starts when the instance is started and ends when the process stops/killed.

each session stats is saved into `{{INSTANCE_DIR}}/stats_v1.json`

the session playtimes are displayed as cards in the quickplay page.

- Resolves: https://github.com/Moulberry/PandoraLauncher/issues/272

### DIFF
--- a/crates/frontend/locales/locales.yml
+++ b/crates/frontend/locales/locales.yml
@@ -338,6 +338,10 @@ instance:
     en: Worlds
   servers:
     en: Servers
+  total_playtime:
+    en: Total Playtime
+  current_session:
+    en: Current Session
   vanilla:
     en: Vanilla
   unable_to_find:
--- a/crates/frontend/src/entity/instance.rs
+++ b/crates/frontend/src/entity/instance.rs
@@ -1,7 +1,7 @@
 use std::{path::Path, sync::Arc};
 
 use bridge::{
-    instance::{InstanceContentSummary, InstanceID, InstanceServerSummary, InstanceStatus, InstanceWorldSummary},
+    instance::{InstanceContentSummary, InstanceID, InstancePlaytime, InstanceServerSummary, InstanceStatus, InstanceWorldSummary},
     message::BridgeDataLoadState,
 };
 use gpui::{prelude::*, *};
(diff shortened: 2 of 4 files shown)


Moulberry:
The reason why I haven't done this yet is because I'm considering how to do this.
I'm not sure what prism does, but the problem with this approach is that the stats won't be accurately tracked if someone closes the launcher.
I'm considering instead making the playtime tracked by the game itself, by having the wrapper (https://github.com/Moulberry/PandoraLauncher/blob/master/wrapper/com/moulberry/pandora/LaunchWrapper.java) spawn a thread that periodically adds 1 to a text file every minute (or something along those lines)

ruri4:
> I'm considering instead making the playtime tracked by the game itself, by having the wrapper

that sounds so impractical and it's too heavy for a simple problem.
"launcher-side tracking is inaccurate if the launcher closes" is true, but tracking the playtime using a wrapper doesn't necessarily follow and is architecturally expensive.

Moulberry:
Is it expensive? Scheduling something to run once every minute to write to a file seems pretty unimpactful. 

There are way more expensive things that launchers do from within the game e.g. discord rpc

CTRL-ALT-OP:
> I'm considering instead making the playtime tracked by the game itself, by having the wrapper (https://github.com/Moulberry/PandoraLauncher/blob/master/wrapper/com/moulberry/pandora/LaunchWrapper.java) spawn a thread that periodically adds 1 to a text file every minute (or something along those lines)

Not entirely sure how these wrappers work, but could you not hook the wrapper or thread to detect when it is being closed/joined and record the timestamp in a text file? And record the timestamp the instance was opened through the launcher, the time played is the difference between the 2 timestamps.
This could also allow for some cool stuff later on, like tracking which instance you usually play at a specific time for stats or sorting.

Moulberry:
> Not entirely sure how these wrappers work, but could you not hook the wrapper or thread to detect when it is being closed/joined

The closest thing would be Runtime.getRuntime().addShutdownHook(), but these are incredibly unreliable

Moulberry:
Thank you for your PR.
I decided that I'll fix the issue with the launcher being closed by daemonizing the backend.
In the future please make sure formatting changes are minimal

PR #279, opened 2026-03-16
Playtime tracking on the instance page and individual instances

bllryy:
- Playtime will also transfer over from Prismlauncher instances
- instances page is now wraped in a div().relative() so the overlay can be absolutely positioned
- scrolling is internal now so the box in the bottom right  stays pinned
- #272 
Pandora


bllryy:
duplicate of #278

Moulberry:
The other PR was merged, sorry

PR #280, opened 2026-03-16
fix: improve instance relocation

notruri:
### Summary

- reworked relocation logic
- the relocate action now uses a folder picker instead of a save-style path prompt (Windows)
- relocation now treats the selected folder as the exact target instance root
- backend relocation no longer relies on rename for cross-drive moves

### DIFF
--- a/crates/backend/src/backend_handler.rs
+++ b/crates/backend/src/backend_handler.rs
@@ -1519,7 +1519,12 @@ impl BackendState {
 
                         self.file_watching.write().unwatch_target(crate::WatchTarget::InstanceDir { id });
 
-                        _ = std::fs::remove_file(&instance.root_path);
+                        if let Err(err) = std::fs::remove_file(&instance.root_path) {
+                            self.file_watching.write().watch_filesystem(instance.root_path.clone(), crate::WatchTarget::InstanceDir { id });
+                            log::error!("Error while deleting link to moved instance: {err:?}");
+                            self.send.send_error(format!("Error while deleting link to moved instance: {err}"));
+                            return;
+                        }
 
                         if is_normal_instance_folder {
                             if let Some(err) = cleanup_moved_instance_source(&target) {


Moulberry:
Did you use AI to write this? I don't trust AI to do the correct thing when it comes to extremely important filesystem operations.

For example this line seems extremely dangerous and could result in instance files being deleted if the move partially completes: <https://github.com/Moulberry/PandoraLauncher/pull/280/changes#diff-bec1f1f53f9f6788bc20fbce614fbb5bbdb4c62f081f29a0576d3e8b92daa8dfR2058>

ruri4:
it was reviewed by an AI but i did it myself, but yeah that's a problem, I'll try to figure out a fix for that

Moulberry:
> it was reviewed by an AI but i did it myself, but yeah that's a problem, I'll try to figure out a fix for that

Perhaps fs_extra::dir::move_dir should only be used if std::fs::rename fails with a specific error indicating that a cross-drive move doesn't work? Either way std::fs::rename should definitely be preferred

Could you show screenshots comparing the before/after of the folder selection prompt on Windows?

ruri4:
### before

{7461E285-9CB2-4D83-BD82-7AA79173DC83}

### after

{3884ECD3-8F9F-4C79-B92E-4B2AA7AF3457}

as you can see, before it uses the file selection instead of folder

Moulberry:
Hmm the new one seems problematic because you need to create the folder beforehand, no?
I think the old one is a lot clearer, even if it might not let you select an existing folder.

Are there any other apps that do something similar with picking a new folder? Which one do they use?

ruri4:
yeah, you would need to create a folder beforehand but it's the about the context rather.

most users expect that to be the instance root rather than instance root storage.

> Are there any other apps that do something similar with picking a new folder? Which one do they use?

they usually prefer where the root is rather than where to store the root

Moulberry:
Hm okay yes I agree the new one is better, maybe if "Select relocation folder" is changed to "Select empty folder" it would be even clearer?

ruri4:
so the new idea for cross filesystem is this:

it's 2 stages move.

stage 1:

1. copy source into a fresh temporary directory near the destination
2. rename the temp into destination

stage 2:

3. remove source
4. link the source into destination

ruri4:
> Hm okay yes I agree the new one is better, maybe if "Select relocation folder" is changed to "Select empty folder" it would be even clearer?

yeah i agree, ill do that as well

Moulberry:
> so the new idea for cross filesystem is this

Sounds good

ruri4:
i tested relocation on Windows and these works (managed = `instances` folder):
- managed to managed
- managed to external
- external to external (internally linked)
- external to managed

just needed more testing especially on linux side.

crates/backend/src/backend_handler.rs
```diff
@@ -1382,11 +1382,42 @@ impl BackendState {
...
+                };
+
+                if path == *instance_root || path == *resolved_instance_root {
+                    self.send.send_info("Instance is already in that location");
                     return;
                 }
 
+                if path.exists() {
+                    if !path.is_dir() {
+                        self.send.send_warning("Cannot relocate instance: path already exists");
```
Moulberry, line 1403: Maybe this message should be more accurate: "path is not a folder"

crates/backend/src/backend_handler.rs
```diff
@@ -1416,74 +1447,154 @@ impl BackendState {
                     }
 
                     #[cfg(windows)]
-                    if let Ok(target) = junction::get_target(&instance.root_path) {
-                        if let Err(err) = std::fs::rename(&target, &path) {
+                    if junction::exists(&instance.root_path).unwrap_or(false) {
+                        let target = match junction::get_target(&instance.root_path) {
```
Moulberry, line 1451: Can you explain the reasoning behind this?
Looking at the implementation of junction, it seems that get_target implies exists. There is never a situation (barring race conditions) where exists returns true while get_target errors.

See https://docs.rs/junction/1.4.2/src/junction/internals.rs.html#115-153

crates/backend/src/backend_handler.rs
```diff
@@ -1416,74 +1447,154 @@ impl BackendState {
...
+                        self.file_watching.write().unwatch_target(crate::WatchTarget::InstanceDir { id });
+
+                        if let Err(err) = junction::delete(&instance.root_path) {
+                            self.file_watching.write().watch_filesystem(instance.root_path.clone(), crate::WatchTarget::InstanceDir { id });
+                            log::error!("Error while deleting junction to moved instance: {err:?}");
+                            self.send.send_error(format!("Error while deleting junction to moved instance: {err}"));
+                            return;
+                        }
+                        
+                        // removing junction doesnt remove the folder for some reason
```
Moulberry, line 1475: I need a very good explanation for why this code is here.
I don't like the look of `remove_dir(&instance.root_path)` unless there's a very good reason for it

crates/backend/src/backend_handler.rs
```diff
@@ -1416,74 +1447,154 @@ impl BackendState {
...
+                        if instance.root_path.is_dir() {
+                            let _ = std::fs::remove_dir(&instance.root_path);
+                        }
+
+                        if is_normal_instance_folder {
+                            if let Some(err) = cleanup_moved_instance_source(&target) {
```
Moulberry, line 1481: I need an extremely good reason why we are deleting the target files here.
I get that the files should have already been moved, but deleting files is extremely dangerous and requires a lot of justification.
If you look at the original code, I never just straight up call remove_dir_all **anywhere**

crates/backend/src/backend_handler.rs
```diff
@@ -1416,74 +1447,154 @@ impl BackendState {
...
+                            instance.rewatch_directories(&mut *file_watching);
+                            file_watching.watch_filesystem(path.clone().into(), crate::WatchTarget::InstanceDir { id });
+                            drop(file_watching);
 
-                        if !is_normal_instance_folder {
-                            if let Err(err) = junction::create(&path, &instance.root_path) {
+                            self.send.send(instance.create_modify_message());
+                            return;
+                        } else {
+                            if let Err(err) = relink_instance_root(&instance.root_path, &target, &path) {
```
Moulberry, line 1494: The whole relink_instance_root change is faulty.
The old target is invalid. We already moved the files.
This should simply be junction::create

crates/backend/src/backend_handler.rs
```diff
@@ -1416,74 +1447,154 @@ impl BackendState {
...
+                                self.file_watching.write().watch_filesystem(instance.root_path.clone(), crate::WatchTarget::InstanceDir { id });
                                 log::error!("Error while creating junction to moved instance: {err:?}");
                                 self.send.send_error(format!("Error while creating junction to moved instance: {err}"));
                                 return;
                             }
+
+                            self.file_watching.write().watch_filesystem(instance.root_path.clone(), crate::WatchTarget::InstanceDir { id });
+
+                            if let Some(err) = cleanup_moved_instance_source(&target) {
```
Moulberry, line 1503: !!!!!!!!!!!

crates/backend/src/backend_handler.rs
```diff
@@ -1416,74 +1447,154 @@ impl BackendState {
...
-                                return;
+                        if let Err(err) = std::fs::remove_file(&instance.root_path) {
+                            self.file_watching.write().watch_filesystem(instance.root_path.clone(), crate::WatchTarget::InstanceDir { id });
+                            log::error!("Error while deleting link to moved instance: {err:?}");
+                            self.send.send_error(format!("Error while deleting link to moved instance: {err}"));
+                            return;
+                        }
+
+                        if is_normal_instance_folder {
+                            if let Some(err) = cleanup_moved_instance_source(&target) {
```
Moulberry, line 1530: !!!!!!!!!!!!!!

crates/backend/src/backend_handler.rs
```diff
@@ -1416,74 +1447,154 @@ impl BackendState {
...
+                            instance.on_root_renamed(self, &path);
+                            let mut file_watching = self.file_watching.write();
+                            instance.rewatch_directories(&mut *file_watching);
+                            file_watching.watch_filesystem(path.clone().into(), crate::WatchTarget::InstanceDir { id });
+                            drop(file_watching);
+
+                            self.send.send(instance.create_modify_message());
+                            return;
+                        } else {
+                            if let Err(err) = relink_instance_root(&instance.root_path, &target, &path) {
```
Moulberry, line 1543: Your refactor of relink_instance_root has now broken the behaviour of this method. Before, if there was a symlink it would recreate a symlink.
Now, when there is a symlink it creates a junction point.
Once again, your relink_instance_root refactor is completely faulty.

crates/backend/src/backend_handler.rs
```diff
@@ -1416,74 +1447,154 @@ impl BackendState {
...
                                 log::error!("Error while linking to moved instance: {err:?}");
                                 self.send.send_error(format!("Error while linking to moved instance: {err}"));
                                 return;
                             }
-                            #[cfg(not(any(unix, windows)))]
-                            compile_error!("Unsupported platform");
+
+                            self.file_watching.write().watch_filesystem(instance.root_path.clone(), crate::WatchTarget::InstanceDir { id });
+
+                            if let Some(err) = cleanup_moved_instance_source(&target) {
```
Moulberry, line 1552: !!!!!!!!!!!!!!!!!!!

crates/backend/src/backend_handler.rs
```diff
@@ -1416,74 +1447,154 @@ impl BackendState {
...
                     }
 
-                    if !is_normal_instance_folder {
-                        #[cfg(unix)]
-                        if let Err(err) = std::os::unix::fs::symlink(&path, &instance.root_path) {
-                            log::error!("Error while linking to moved instance: {err:?}");
-                            self.send.send_error(format!("Error while linking to moved instance: {err}"));
-                            return;
+                    if is_normal_instance_folder {
+                        if let Some(err) = cleanup_moved_instance_source(&source_path) {
```
Moulberry, line 1570: !!!!!!!!!!!!!!!!!!!!!

crates/backend/src/backend_handler.rs
```diff
@@ -2027,6 +2138,157 @@ impl BackendState {
     }
 }
 
+fn move_instance_directory(src: &Path, dest: &Path) -> Result<(), Arc<str>> {
```
Moulberry, line 2141: This method is still bad.
As I told you, we need to try std::fs::rename first.
std::fs::rename is an atomic move and is 1000x better than all this nonsense that you do with temporary directories and recursive copies.
ONLY and I mean ONLY if it is IMPOSSIBLE to do the atomic move should we resort to trying all this other stuff

crates/backend/src/backend_handler.rs
```diff
@@ -2027,6 +2138,157 @@ impl BackendState {
...
+    dest: &Path,
+    temp_dir: &Path,
+) -> Result<(), Arc<str>> {
+    // first copy the contents into the temp dir
+    let options = fs_extra::dir::CopyOptions::default().content_only(true);
+    fs_extra::dir::copy(src, temp_dir, &options)
+        .map_err(|err| Arc::<str>::from(format!("Unable to copy instance files: {err}")))?;
+
+    // remove destination first if it exists
+    let dest_exists = dest.exists();
```
Moulberry, line 2208: toctou, you dont need to check exist and then remove. Just do `_ = std::fs::remove_dir()` directly

crates/backend/src/backend_handler.rs
```diff
@@ -2027,6 +2138,157 @@ impl BackendState {
...
+        ).into());
+    }
+
+    Ok(())
+}
+
+fn replace_source_directory_with_link(src: &Path, dest: &Path) -> Result<Option<Arc<str>>, Arc<str>> {
+    let parent = src.parent().ok_or_else(|| Arc::<str>::from("Unable to replace instance directory with a link: source has no parent directory"))?;
+    let source_temp = prepare_move_temp_dir(parent, src)?;
+    std::fs::rename(src, &source_temp)
```
Moulberry, line 2268: What's the point of this source_temp stuff?

Moulberry:
I'm sorry but moving around files is **extremely** critical functionality, and very careful care should be taken to ensure the files are not deleted. The move failing with the files intact is 1000x preferable to the move failing and the files getting obliterated. The rampant use of remove_dir_all and other shenanigans in this PR makes me extremely wary of accepting it.

When it comes to important stuff like this, care should be taken to do **as little as possible**, which is why the original code was only 100 lines of code. This PR adds close to 300, and I wouldn't be surprised if there is a subtle toctou bug that results in the instance data being lost.

Moulberry:
There's so much that is unnecessary, for example doing instance.on_root_renamed isn't needed because it's handled by the filesystem listener. Doing it like this only introduces race conditions.

Most of the additional error handling is also unnecessary, because if the `std::fs::remove_file(&instance.root_path)` fails and the file still exists, then the following code that creates the new symlink (`std::os::unix::fs::symlink(&path, &instance.root_path)`) will fail. The only thing that adding extra error handling does is increase the surface area and risk for toctou bugs.

Really, the only changes here that I see are needed is the unwatching/rewatching (though even then I'm not sure if it's needed, since the watch should be on the underlying canonical dir) and the **fallback** movement that involves copying files, which should be a fallback instead of replacing std::fs::rename

Moulberry:
Ok I'm sorry again but I have to close this. I just don't feel comfortable with these changes and the risk they carry to user's data. These matters are extremely important to me, because a mistake in the UI can be fixed, but a mistake in the file handling cannot

crates/backend/src/backend_handler.rs
ruri4: removing the junction doesn't remove the folder itself, that's why it has to be explicitly deleted

PR #285, opened 2026-03-16
Add languages switch + Polish translations

Rajdacz:
Extracted remaining hardcoded strings to locales, added a language switcher in settings, and included Polish translations.

I'm considering using LLM to generate translations for a few popular languages, but I know a lot of people don't like using it for that. Should I do it, or do you prefer to wait for people to PR new languages manually?

### DIFF
--- a/crates/frontend/locales/locales.yml
+++ b/crates/frontend/locales/locales.yml
@@ -958,30 +958,6 @@ instance:
         en: Installed manually - cannot automatically update
         pl: Zainstalowano ręcznie - nie można aktualizować automatycznie
 
-  # Import Feature
-  import:
-    label:
-      en: Import
-    from:
-      en: Import from %{launcher}
-      pl: Importuj z %{launcher}
-    modrinth_pack:
-      label:
-        en: Import Modrinth Pack (.mrpack)
-        pl: Importuj Paczkę Modrinth (.mrpack)
-      select:
-        en: Select Modrinth Pack
-        pl: Wybierz Paczkę Modrinth
-    accounts:
-      en: Import Accounts
-      pl: Importuj Konta
-    instances:
-      en: Import Instances
-      pl: Importuj Instancje
-    error:
-      en: Error importing
-      pl: Błąd importowania
-
   # Sync Feature
   sync:
     label:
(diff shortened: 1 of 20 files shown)


linuxchr:
I dont think putting all translations in a single file is the right way

Rajdacz:
> I dont think putting all translations in a single file is the right way

I think the same, but Moulberry made locales.yml file instead of en.yml, so I think he did that on purpose. If not, I can change that

linuxchr:
And on the topic of LLM translations, I would rather wait for PRs. I would offer myself for german translations

crates/frontend/locales/locales.yml
```diff
@@ -201,585 +261,951 @@ modrinth:
...
+          pl: Pobierz aktualizację z Curseforge
       error:
         en: Error updating mod
+        pl: Błąd aktualizacji modyfikacji
       installed_manually:
         en: Installed manually - cannot automatically update
+        pl: Zainstalowano ręcznie - nie można aktualizować automatycznie
+
+  # Import Feature
+  import:
```
dragmine149: Personally, i would put this top-level instead of under `instance` kinda like i've done in #286.
Just because we do *,as shown further down,* import both instances and accounts. Additionally, `instance.import.instances` seems a bit excessive...

Rajdacz: Moving import to top-level is a good point, but what do you mean about `instance.import.instances`? It is used in the checkbox while importing

dragmine149: The use of it is not excessive, the naming is.
You're basically duplicating a section of the name. as either `import.instance` or `instance.import` are both valid for what it's actually doing.

Personally, I don't think there should be any cases where `{x}.{y}.{x}` exists. As you could probably rewrite it into either `{x}.{y}` or `{y}.{x}` or smth else. I understand there are cases where `{x}.{y}.{x}` is valid, just most of the time it's cleaner to rewrite said case.

Rajdacz: Oh, I see what you mean. Anyway, I moved it outside of `instance`, so now it's just `import.instances`

Moulberry:
e434e92436864cc08d70cbeb541b45883f001361 reworked how translations are done, so it is now ready to be translated to other languages

Issue #276, opened 2026-03-15
[FEATURE REQUEST] Add icons on exsisting instances


disconnectedgrinders:
I would like to be able to change the icon for the instance on the instance page by clicking the icon. Add a hover effect  of a pencil or icon that indicates you are about to open the icon menu. Then have this icon be used when a shortcut is created aswell.
<img width="1280" height="476" alt="Image" src="https://github.com/user-attachments/assets/657311c1-02a9-4054-b8d3-d725edc6efcf" />

disconnectedgrinders:
Yes both way work, I changed my issue as i figured it would be faster to not have to open settings

Moulberry:
Added by #297

PR #297, opened 2026-03-16
Add changing icon to instances settings

Rajdacz:
Relates #276 (issue was edited, now it doesn't close it entirely)

I also tried putting it as an icon next to the instance name, but it looked unreadable

icon


Moulberry:
Thanks for your PR

PR #299, opened 2026-03-17
Update importers to import the preferred account as well

dragmine149:
Simple changes to import, just includes account in config where possible. Continuation of #249  in a way


Moulberry:
Thank you for your PR

PR #301, opened 2026-03-17
Add clear and copy buttons for console

Antsahw:
just added clear and copy button to the console


Antsahw:
dont look at the Update instance.rs idk what i was cooking like it was giving a warnings when i compiled it and i was trying to solve it. dk if its correct to do like that

Antsahw:
ah warnings were already existing, not from my changes

Antsahw:
is alright that i "fixed" them?

Antsahw:
i will unfix it. its someones else work

Antsahw:
it was little mess but now its fine

PR #306, opened 2026-03-18
Added remove skin button and ability to delete skin from skin directory

henbotb:
Adds a trash can icon on the bottom left of every skin option in the skin menu to delete the skin.

With this implementation, however, the click event on the skin png is also registered, meaning that it registers it as the selected skin whilst removing it, which technically doesn't matter but would be nice to fix for QOL purposes (and also minor performance given the skin widget wouldn't need to reload). I read a few ways to suspend the click event / stop click propagation, but didn't find much documentation on them.

This would also help resolve another minor visual bug where if you delete a skin when it's in a selected state, it removes the selection as a whole.

### DIFF
--- a/crates/frontend/src/pages/skins_page.rs
+++ b/crates/frontend/src/pages/skins_page.rs
@@ -11,7 +11,7 @@ use rustc_hash::FxHashMap;
 use schema::minecraft_profile::{SkinState, SkinVariant};
 use uuid::Uuid;
 use crate::{
-    component::{player_model_widget::PlayerModelWidget, shrinking_text::ShrinkingText}, data_asset_loader::DataAssetLoader, entity::{DataEntities, account::AccountExt}, icon::PandoraIcon, interface_config::InterfaceConfig, pages::page::Page, png_render_cache::ImageTransformation, ts
+    component::{player_model_widget::PlayerModelWidget, shrinking_text::ShrinkingText}, data_asset_loader::DataAssetLoader, entity::{DataEntities, account::AccountExt}, icon::PandoraIcon, interface_config::InterfaceConfig, pages::page::Page, png_render_cache::ImageTransformation, skin_renderer::{self, determine_skin_variant}, ts
 };
 
 pub struct SkinsPage {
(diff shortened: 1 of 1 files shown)


crates/backend/src/skin_manager.rs
```diff
@@ -54,6 +54,42 @@ impl SkinManager {
...
+
+        let Ok(_) = std::fs::remove_file(skin_path) else {
+            log::warn!("Unable to delete skin");
+            return;
+        };
+
+    }
+
+    fn get_path_from_skin(backend: &BackendState, skin: Arc<[u8]>) -> Option<PathBuf> {
+        let Ok(read_dir) = std::fs::read_dir(&backend.directories.skin_library_dir) else {
```
Moulberry: Scanning the whole directory like this seems not ideal, perhaps the skin manager should maintain a map from skin -> path

Moulberry (changes requested):

crates/backend/src/skin_manager.rs
henbotb: Totally agree, and had the same thought while programming this. I ended up opting for this solution as the it's a fairly low-access function and I didn't see an obvious way to implement something like a map without having to add an extra parameter to a bunch of closures. That said, I will look into something less costly. 

If you have any ideas for the structure, let me know.

henbotb:
I wish that I could say these 12 lines of code were the product of 20 minutes of work, but I went through a few different iterations of this system which all seemed inelegant or had clunky iterator logic going on. 

The idea initially was to group the paths with the existing skin bytes in the skin library, but that ran into a few different issues with how other functions interact with it that weren't as simple as adding `.0` to the end of paths since the active skin isn't actually associated with a path when its read in from Mojang in the first place. I suspect there might still be a way to do this, but I was having little success.

This is still subjected to the aforementioned issues where the click event doesn't get consumed so it still changes to the before deleting since the click is being read in in multiple places for the same container, however. If you happen to know any way to stop propagating the click event that'd probably resolve this issue, but looking stuff up about it only results in automatic AI responses hallucinating code (seemingly) and no/very little actual documentation on the matter.

Moulberry:
The way to stop propagation is usually with `cx.stop_propagation()` and `window.prevent_default()`. You're probably missing `window.prevent_default()`, the docs for it say that it's designed to prevent the parent being focused

Moulberry:
Please fix the merge conflicts

henbotb:
> Please fix the merge conflicts

I apologize if this has been a messy PR as it's my first, and I'm not super familiar with Rust, but that should have worked. Let me know if it didn't/if there's anything else expected.

Moulberry (approved):

Moulberry:
Thank you for your PR

PR #309, opened 2026-03-18
Add support for importing instances from curseforge

dragmine149:
Probably time this got introduced now...

Basically, just copied over most of the code from the `ATLauncher` import system with changing some variables here and there... (maybe at some point we turn this into a more reuse-old-code system rather than copy-paste?)

***annoyingly***, the default curseforge dir is `~/Documents/curseforge` instead of `~/.local/share/curseforge` (for linux) so the bridge got modified to take all directories from `directories::base_dirs` instead of just the data dir.

The rest of curseforge config seems to be under the normal electron folder of `~/.config/CurseForge` (yet another weird path) which although not impossible to support importing from, does take a lot more work.

The things i'm not a big fan of are:
- How the icon importing is handled
- How the mod loader is recieved (split a string)

They work, i've tested them but from my experience of other code in this codebase, it just doesn't fully feel right...

### DIFF
--- a/crates/backend/src/launcher_import/curseforge.rs
+++ b/crates/backend/src/launcher_import/curseforge.rs
@@ -77,7 +77,6 @@ pub fn import_instances_from_curseforge(backend: &BackendState, import_job: &Imp
     let mut to_import = Vec::new();
 
     for folder in import_job.paths.iter() {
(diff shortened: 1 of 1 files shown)


Moulberry:
Thank you for your PR

PR #322, opened 2026-03-20
Replace skins textures with 3D preview

### LINKED ISSUE #262: Show skin previews as 3D rendered models instead of raw PNG
## Problem
Skins in the library/selector currently display as raw PNG image files, making it difficult to preview how they actually look when applied to the player model

## Expected Behavior
Display skins as 3D rendered previews (like the view when skin is selected) for better visualization before application

## Current Behavior
Raw PNG texture files shown in list format.

Image

Rajdacz:
Closes #262 
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,6 +631,7 @@ dependencies = [
  "base64 0.22.1",
  "bridge",
  "chrono",
+ "command",
  "crc32fast",
  "directories",
  "enum-map",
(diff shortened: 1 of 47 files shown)


Moulberry:
I'd prefer for this to be an option. I still think being able to see the actual png like you might see in the file browser is helpful

Moulberry:
Thank you for your PR

PR #327, opened 2026-03-21
Add steal skin feature

### LINKED ISSUE #325: [Feature Request] Copy skin from username in skins tab
This feature would be really neat as its basically what essentails allows you to do but you can now do it directly in the launcher without having to download essentials which is amazing.

Rajdacz:
Closes #325 
image

### DIFF
--- a/crates/backend/src/backend_handler.rs
+++ b/crates/backend/src/backend_handler.rs
@@ -238,16 +238,32 @@ impl BackendState {
                     return;
                 };
 
-                if instance.processes.is_empty() {
+                if instance.processes.is_empty() && instance.closing_processes.is_empty() {
                     self.send.send_error("Can't kill instance, instance wasn't running");
                     return;
                 }
 
-                for mut process in instance.processes.drain(..) {
+                for (process, _) in instance.closing_processes.drain(..) {
                     let result = process.kill();
-                    if result.is_err() {
+
+                    if let Err(err) = result {
                         self.send.send_error("Failed to kill instance");
-                        log::error!("Failed to kill instance: {:?}", result.unwrap_err());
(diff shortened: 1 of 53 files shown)


crates/backend/src/backend_handler.rs
```diff
@@ -1783,13 +1783,152 @@ impl BackendState {
...
+                    "https://api.mojang.com/minecraft/profile/lookup/name/{}",
+                    username
+                );
+                let response = match self.http_client.get(&lookup_url).send().await {
+                    Ok(r) => r,
+                    Err(err) => {
+                        log::error!("StealSkin: failed to reach Mojang API: {:?}", err);
+                        self.send.send_error(format!("Could not find player '{}'", username));
+                        return;
+                    }
```
Moulberry, line 1816: Check the response code for 404, that should say "Player not found"
Any response code other than 200 should be considered an error
All other messages should say something other than "Player not found", like failed to request, failed to read, failed to deserialize, etc.

crates/backend/src/backend_handler.rs
```diff
@@ -1783,13 +1783,152 @@ impl BackendState {
                     self.send.send_error("Error while saving skin, see logs for more details");
                 }
             },
+            MessageToBackend::StealPlayerSkin { username } => {
```
Moulberry: Can we change the wording? "Steal skin" feels very joke-y

Rajdacz: Sure, I changed it to "Copy from player"

crates/backend/src/backend_handler.rs
Rajdacz: Done

Moulberry:
Thank you for your PR

Issue #320, opened 2026-03-20
[Feature request] Allow servers to be reorderd form the serverlist in the launcher


disconnectedgrinders:
Its not a big deal but i just randomly had the need for it and makes things sligthly faster in some situations

PR #343, opened 2026-03-25
Add server list reorder and persist to servers.dat

5sig2:
Added up and down controls to reorder servers in instance quickplay list and persist the order back to the `servers.dat` so minecraft and the launcher stay in sync
Suppress refreshing from self-triggered file changes to avoid Icon Flicker

Edit: closes https://github.com/Moulberry/PandoraLauncher/issues/320

### DIFF
--- a/crates/backend/src/backend.rs
+++ b/crates/backend/src/backend.rs
@@ -147,6 +147,7 @@ pub fn start(launcher_dir: PathBuf, send: FrontendHandle, self_handle: BackendHa
         cached_minecraft_profiles: Default::default(),
         skin_manager: Default::default(),
         server_list_pinger: Arc::new(ServerListPinger::new()),
+        server_dat_ignore: Default::default(),
     };
 
     log::debug!("Doing initial backend load");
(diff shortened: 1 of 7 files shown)


Moulberry:
In Pandora, the filesystem should be the single source of truth. Suppressing filesystem changes isn't the right way to do this, since there could be a race condition if the server.dat is being manipulated by another program simultaneously.

The way I would do it is to set a flag internally, and if a second change occurs before the filesystem signals us then we reload the content from disk directly. (You can look at the persistent.rs for an example of this pattern)

Getting this right is not easy, so I won't ask you to do it. I'll merge these changes as-is and then modify it myself to be more robust. (Merge will be a little bit delayed because 26.1 just came out and I'm working on a big feature right now in Pandora)

5sig2:
> In Pandora, the filesystem should be the single source of truth. Suppressing filesystem changes isn't the right way to do this, since there could be a race condition if the server.dat is being manipulated by another program simultaneously.
> 
> The way I would do it is to set a flag internally, and if a second change occurs before the filesystem signals us then we reload the content from disk directly. (You can look at the persistent.rs for an example of this pattern)
> 
> Getting this right is not easy, so I won't ask you to do it. I'll merge these changes as-is and then modify it myself to be more robust. (Merge will be a little bit delayed because 26.1 just came out and I'm working on a big feature right now in Pandora)

Sounds good

disconnectedgrinders:
> (Merge will be a little bit delayed because 26.1 just came out and I'm working on a big feature right now in Pandora)

What are you working on for 26.1 in pandora?

Moulberry:
Thank you for your PR

PR #350, opened 2026-03-27
Add dependency selector for modrinth & curseforge installs

### LINKED ISSUE #346: [Feature request] Download mod dependencies when adding from Modrinth
prism launcher for example asks if u want to download the dependencies when u are downloading from modrinth

Image

5sig2:
Added a dependency selector when installing a mod with dependencies and sub dependencies. Dependencies already installed in the target instance are not shown. Sub-dependencies are resolved recursively in the UI.

“Install Latest” is unchanged and still installs immediately with the existing auto-dependency behavior.

Closes https://github.com/Moulberry/PandoraLauncher/issues/346

image

### DIFF
--- a/crates/frontend/src/modals/curseforge_install.rs
+++ b/crates/frontend/src/modals/curseforge_install.rs
@@ -4,12 +4,12 @@ use bridge::{install::{ContentDownload, ContentInstall, ContentInstallFile, Inst
 use enumset::EnumSet;
 use gpui::{prelude::*, *};
 use gpui_component::{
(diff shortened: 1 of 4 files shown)


Moulberry:
Why is the open_latest/open_latest_from_entity code still here if "Install Latest" is unchanged?

5sig2:
> Why is the open_latest/open_latest_from_entity code still here if "Install Latest" is unchanged?

Originally I had coded in open_latest/open_latest_from_entity so that install latest would run  through the same path that supports dependency selection, but in our discord conversation you told me it install latest should just do it without asking for dependency selection so I removed the dependency selection UI, the internal logic is the same, but I forgot to revert it back to inline install logic in the page

PR #366, opened 2026-04-01
Added instance exporting, zip, modrinth and curseforge

### LINKED ISSUE #363: [Feature Req] Allow exporting packs
I would like to use pandora to make modpacks to my friends and such. So I need to export to both MrPack and ZIP. It would also be useful if you have multible computers (which i have).

5sig2:
Adds instance exporting to the instance settings page, with ZIP, mrpack and curgeforge pack formats. Took inspiration from prism's export system.

Closes https://github.com/Moulberry/PandoraLauncher/issues/363


Moulberry:
What does "took inspiration" mean? Because Prism's code is GPL

5sig2:
> What does "took inspiration" mean? Because Prism's code is GPL

I studied how prism implemented the mrpack export, then did a independent implementation based on the knowledge I gathered from that and the modrinth documentation, from what I understand about both licenses this should be fine, as it exempts it from being considered a derivative work

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+    let files = collect_files(
+        &instance.root_path,
+        options,
+        &instance.sync_targets,
+        &backend.directories.synced_dir,
+        modal_action,
+    )?;
+    tracker.set_total(files.len());
+    tracker.notify();
+    tracker.set_finished(ProgressTrackerFinishType::Normal);
```
Moulberry, line 167: What's the point in having this tracker if you set the total right before finishing it?

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+    let mut exclude = HashSet::new();
+    for resolved_file in &resolved {
+        exclude.insert(resolved_file.rel_path.clone());
+    }
+
+    let manifest_json = build_curseforge_manifest(instance, options, &resolved)?;
+    let modlist_html = build_curseforge_modlist(&resolved);
+    let extra_files = vec![
+        ("manifest.json".to_string(), manifest_json),
+        ("modlist.html".to_string(), modlist_html),
```
Moulberry, line 263: What is this file? I just checked a normal curseforge modpack (ATM10) and it doesn't contain this file

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+        for target in sync_targets.files.iter() {
+            let target = target.as_ref().replace('\\', "/");
+            if candidate == target {
+                return true;
+            }
+        }
+    }
+
+    false
+}
```
Moulberry, line 396: Sync targets should always be relative to the .minecraft folder, so you shouldn't need to scan multiple "candidates". Just take path relative to dot_minecraft_path.

Instead of comparing strings, you should convert the string target into a Path and then compare the paths (or use path.starts_with), this should make the handling much more reliable and avoids needing to replace backslashes and stuff.

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+
+fn should_skip(rel: &Path, options: &ExportOptions) -> bool {
+    let Some(component) = rel.components().next() else {
+        return false;
+    };
+    let name = match component {
+        std::path::Component::Normal(name) => name.to_string_lossy(),
```
Moulberry: This is a minor nitpick, but it's relevant in other parts of the code as well, it's better to use as_encoded_bytes and then compare it to bytes e.g. `b"crash-reports"`. This avoids needing to allocate an extra string

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+    }
+    if !options.include_mods && name == "mods" {
+        return true;
+    }
+    if !options.include_resourcepacks && name == "resourcepacks" {
+        return true;
+    }
+    if !options.include_configs && name == "config" {
+        return true;
+    }
```
Moulberry, line 448: A match on `name` might be more readable

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+
+async fn resolve_modrinth_files(
+    backend: &BackendState,
+    instance: &ExportInstanceData,
+    options: &ExportOptions,
+    files: &[ExportFile],
+    modal_action: &ModalAction,
+    tracker: &ProgressTracker,
+) -> Result<Vec<ModrinthResolvedFile>, ExportError> {
+    if !options.include_mods {
```
Moulberry: You can also upload resourcepacks to modrinth

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+
+        let (sha1_hex, sha512_hex, size) = compute_hashes(&file.abs, modal_action)?;
+
+        if modrinth_loader == ModrinthLoader::Unknown {
+            continue;
+        }
+
+        let response = backend
+            .meta
+            .fetch(&ModrinthVersionUpdateMetadataItem {
```
Moulberry: Won't this get the updated file instead of the file that the user actually has installed?
We should be using this request instead, and make a single request for all files so we don't get ratelimited: https://docs.modrinth.com/api/operations/versionsfromhashes/

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+
+    let mut files_out = Vec::new();
+    for file in resolved {
+        let mut env = serde_json::Map::new();
+        if file.optional {
+            env.insert("client".into(), serde_json::Value::from("optional"));
+            env.insert("server".into(), serde_json::Value::from("optional"));
+        } else {
+            env.insert("client".into(), serde_json::Value::from("required"));
+            env.insert("server".into(), serde_json::Value::from("required"));
```
Moulberry: This might be problematic. It's going to say that sodium is "required" on the server, for example.
We should either omit the `env` field entirely (it's optional), or get this information from the file's modrinth project page

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+    Ok(resolved)
+}
+
+fn build_modrinth_index(
+    instance: &ExportInstanceData,
+    options: &ExportOptions,
+    resolved: &[ModrinthResolvedFile],
+) -> Result<Vec<u8>, String> {
+    let mut out = serde_json::Map::new();
+    out.insert("formatVersion".into(), serde_json::Value::from(1));
```
Moulberry: It might be better to use a schema (schema::mrpack::ModrinthIndexJson already exists, but might need some additional fields to be added) instead of building the json from raw values

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+        let _ = fs::remove_file(&temp_path);
+    }
+
+    result
+}
+
+fn temp_output_path(output: &Path) -> PathBuf {
+    let mut temp = output.to_path_buf();
+    let ext = output.extension().and_then(OsStr::to_str).unwrap_or("tmp");
+    temp.set_extension(format!("{}.new", ext));
```
Moulberry: You can use PathBuf::add_extension

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+    temp
+}
+
+fn compute_hashes(path: &Path, modal_action: &ModalAction) -> Result<(String, String, u64), ExportError> {
+    let mut file = File::open(path).map_err(|e| e.to_string())?;
+    let mut sha1 = Sha1::new();
+    let mut sha512 = Sha512::new();
+    let mut buffer = [0_u8; 8192];
```
Moulberry: Could probably use a bigger buffer here, just re-use the allocation for all the compute_hashes calls

crates/backend/src/backend_handler.rs
```diff
@@ -92,6 +92,12 @@ impl BackendState {
                     }
                 }
             },
+            MessageToBackend::ExportInstance { id, format, options, output, modal_action } => {
+                let backend = self.clone();
+                tokio::task::spawn(async move {
```
Moulberry, line 100: Doesn't really matter, but for future reference you should just be able to do:
`tokio::task::spawn(crate::export::export_instance(..))`

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+    let Some(component) = rel.components().next() else {
+        return false;
+    };
+    let name = match component {
+        std::path::Component::Normal(name) => name.to_string_lossy(),
+        _ => return false,
+    };
+
+    if !options.include_logs && (name == "logs" || name == "crash-reports") {
+        return true;
```
Moulberry: Any file with the extension `.log` or `.log.gz` should also probably be excluded (e.g. `fabricloader.log`, `hs_err_pid<pid>.log`, and any custom files added by mods)

crates/frontend/src/modals/export_instance.rs
```diff
@@ -0,0 +1,346 @@
...
+                2 => ExportFormat::Curseforge,
+                _ => ExportFormat::Zip,
+            };
+            cx.notify();
+        }
+    }
+
+    fn build_options(&self, cx: &mut App) -> ExportOptions {
+        let modrinth_summary = self.modrinth_summary_input.read(cx).value();
+        let modrinth_summary = if modrinth_summary.trim().is_empty() {
```
Moulberry: As a general rule I prefer trim_ascii over trim

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+    }
+
+    false
+}
+
+fn is_os_junk(rel: &Path) -> bool {
+    let Some(file_name) = rel.file_name().and_then(OsStr::to_str) else {
+        return false;
+    };
+    matches!(file_name, ".DS_Store" | "Thumbs.db" | "thumbs.db")
```
Moulberry: Please also exclude files starting with `.pandora.` (these are temporary files used when loading a modpack on forge for example) and files starting with . and ending in .aux.json (these files store metadata about the installed modpack)

crates/backend/src/export.rs
```diff
@@ -0,0 +1,841 @@
...
+    instance: &ExportInstanceData,
+    options: &ExportOptions,
+    output: &Path,
+    modal_action: &ModalAction,
+) -> Result<(), ExportError> {
+    check_cancel(modal_action)?;
+    let collect_tracker = ProgressTracker::new("Collecting files".into(), backend.send.clone());
+    modal_action.trackers.push(collect_tracker.clone());
+
+    let files = collect_files(
```
Moulberry, line 190: Pandora allows you to load modpacks directly inside an instance. This might cause problems for other launchers though. We should probably detect modpack files inside instances and merge their contents into the modpack we export.

Should be fairly straightforward if including a mrpack when exporting to mrpack, but might require some web requests if including a curseforge modpack inside an mrpack, because we then have to convert all the files

Cargo.toml
```diff
@@ -76,6 +76,7 @@ oauth2 = { version = "5.0.0", features = ["reqwest"] }
 p256 = "0.13.2"
 base64ct = "1.8.0"
 sha2 = "0.10.9"
+zip = "0.6.6"
```
Moulberry: Why is 0.6.6 being used? The latest is 8.5.1

Moulberry (changes requested):

crates/backend/src/export.rs
5sig2: Yeah no, just a slip up by me

crates/backend/src/export.rs
5sig2: From what I understand it's a legacy / optional curseforge pack artifact, so it doesnt tend to show up in modern packs
Example: https://github.com/Divine-Journey-2/Divine-Journey-2

crates/backend/src/export.rs
5sig2: Alright noted, I'll add this to my next refactor commit

5sig2:
> Pandora allows you to load modpacks directly inside an instance. This might cause problems for other launchers though. We should probably detect modpack files inside instances and merge their contents into the modpack we export.
> 
> Should be fairly straightforward if including a mrpack when exporting to mrpack, but might require some web requests if including a curseforge modpack inside an mrpack, because we then have to convert all the files

@Moulberry Did all the requested changes, with the exception the change above, wasn't able to figure it out as easily due to being somewhat busy and not fully understanding how to make this change, if you still want me to implement this, could you just give me more details on how this would really work, as I don't really understand it, if there's any other changes you need me to make let me know.

Moulberry:
It's fine, it's not strictly necessary, and it's a bit complicated to do anyways.
This is probably fine as a first implementation

Moulberry:
Thank you for your PR

PR #383, opened 2026-04-10
[Fix] Use Configured Minecraft Version Instead of Hardcoded "1.21.10" (Closes #377)

### LINKED ISSUE #377: [Bug] Incorrect version number passed to game
The version number passed to the game in https://github.com/Moulberry/PandoraLauncher/blob/master/crates/backend/src/launch/mod.rs#L2374 is hardcoded to `"1.21.10"` it should instead be `&self.configuration.minecraft_version`. This string is used in f3 debug for the game version entry.

fooeyround:
Closes / Implements https://github.com/Moulberry/PandoraLauncher/issues/377


Moulberry:
Thank you for your PR

Issue #375, opened 2026-04-07
[bug] Unable to use Simple Voice CHat


shardmods:
<img width="745" height="56" alt="Image" src="https://github.com/user-attachments/assets/b24257bc-0d36-4852-a9ca-68c6b096da53" />

OS: MacOS Tahoe 26.3.1
Version: Latest

PR #385, opened 2026-04-11
[Feature] MacOS Microphone Permission

fooeyround:
This PR adds an Info.plist and entitlements.plist to add the required `NSMicrophoneUsageDescription` and entitlement for microphone access.

Locally this works for me _after_ I self sign  `PandoraLauncher.app` with my own developer account.
Without signing the app MacOS asks for the permission repeatably before stopping; the microphone does work, though. 

On line 40 of build_mac.sh you delete the app after archiving it.
I believe you might need to sign it first:
```sh
if [[ -n "$CARGO_PACKAGER_SIGN_PRIVATE_KEY" ]]; then
    cargo packager signer sign dist/PandoraLauncher.app
fi
```
The signature for the app seems to be (as of the last release) adhoc. With that configuration the feature didn't seem to work. 
Before this feature could work there may need to be changes to the build process to include a developer signature.
Is there a specific reason why you unset the signature when building? Does cargo-packager not handle signing the way you like by default?

### DIFF
--- a/package/mac/Info.plist
+++ b/package/mac/Info.plist
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSMicrophoneUsageDescription</key>
+	<string>A Minecraft mod wants to access your microphone.</string>
+</dict>
+</plist>
--- a/package/mac/entitlements.plist
+++ b/package/mac/entitlements.plist
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.device.audio-input</key>
+	<true/>
+</dict>
+</plist>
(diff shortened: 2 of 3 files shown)


fooeyround:
When complete this would close https://github.com/Moulberry/PandoraLauncher/issues/375, but it also seems as though some more changes need to be made for `librnnoise` to be able to load. With signing the prompt occurs once and never again. I believe it may require more entitlements, but am unsure.

Moulberry:
> On line 40 of build_mac.sh you delete the app after archiving it.

The .app is not a file, it is a folder

>  I believe you might need to sign it first

The `cargo packager signer sign` command is not for code signing. It generates a .sig file which is used by the auto-updating code to validate that the GitHub releases aren't compromised.

MacOS codesigning is different and I believe it requires a $100/yr apple developer subscription or something

fooeyround:
> The .app is not a file, it is a folder

I am aware, but it is functionally the "App" bundle as MacOS is concerned. `codesign` works on not only binaries but app bundles as well. Sorry for not being clear. Right before archiving is where you would use codesign; but of course it would need to be on a mac with the developer account in the keychain.

> The cargo packager signer sign command is not for code signing. It generates a .sig file which is used by the auto-updating code to validate that the GitHub releases aren't compromised.

Oh, sorry, I was under the assumption it did both, though that makes sense. I am not very familiar with cargo-packager and only looked at the docs briefly on how to pass the `Info.plist` and `entitlements.plist`.
> MacOS codesigning is different and I believe it requires a $100/yr apple developer subscription or something

I am using the macos codesign utility to sign the app bundle. ~I don't believe it requires the paid license, just one tied to an apple account.~ I am using a paid developer subscription, though. I will try and test with the free signing profile.

This is a big reason why this is only a draft. I don't know what you would want to do with this.
It seems to _work_ though (as in, the permission works but it is ephemeral). It would just be nice to find out how to make the permission persistent without the license.

fooeyround:
Sadly, you are right; a paid license is required to do code-signing. This should "work" adhoc though; it seems like it already does, just very _weirdly_.
My next step on this PR is definitely going to be looking into permission persistence with an adhoc signature.

Moulberry:
Going to merge this anyways, another PR with the signing stuff can be made

Moulberry:
Thank you for your PR

PR #444, opened 2026-04-25
fix: silent failure when installing content across different partitions

BluePotatoBP:
In the current implementation, std::fs::hard_link silently faills across volumes (in windows terms) when installing content.

I got ERROR_NOT_SAME_DEVICE when trying to update a modpack because the content library is stored on the C drive (appdata) and i moved the instance to a separate volume (X: drive). This resulted in the old file being deleted before the link attempt so both the old and new files vanish.

The proposed fix implements hard_link_or_copy which determins if a file copy should be made as a fallback in case the content lib isnt on the same volume- but still prefers the hard_link. Yes this is a bandaid and the proper fix would be to expose the content library path but that raises its own set of problems (mostly the data migration and the way it should be handled) so i leave that up to your sanction.


Moulberry:
Thank you for your PR

Issue #432, opened 2026-04-20
[Feature Request] Add an option to delete servers


Swylow1:
Add an option to delete servers inside the Pandora Launcher

shardmods:
Seems like it would be a dupe or at least part of #429

bllryy:
imma see if i can do this

bllryy:
#445

PR #445, opened 2026-04-25
feat: Added the feature to delete servers from the server list

bllryy:
I have an example photo here 
image

### DIFF
--- a/crates/backend/src/backend_handler.rs
+++ b/crates/backend/src/backend_handler.rs
@@ -78,6 +78,9 @@ impl BackendState {
             MessageToBackend::ReorderServers { id, from_index, to_index } => {
                 tokio::task::spawn(Instance::reorder_servers(self.clone(), id, from_index, to_index));
             },
+            MessageToBackend::DeleteServer { id, index } => {
+                tokio::task::spawn(Instance::delete_server(self.clone(), id, index));
+            },
             MessageToBackend::RequestLoadMods { id } => {
                 tokio::task::spawn(Instance::load_content(self.clone(), id, ContentFolder::Mods));
             },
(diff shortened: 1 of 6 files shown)


bllryy:
this is refrence to [https://github.com/Moulberry/PandoraLauncher/issues/432](url)

shardmods:
I fw it

Moulberry:
I'm not 100% sure about this. It takes up a lot of visual space and doesn't seem super useful

disconnectedgrinders:
If you make the server lists drag to reorder instead of buttons it would be less cramped as there would only be the delete button.

Moulberry:
If I'm gonna do drag to reorder then I'd still prefer to not have the delete button there

disconnectedgrinders:
How about this, if you CTRL click any or multiple servers to select them a button in the region where the server text is appears to delete all selected servers.

bllryy:
> How about this, if you CTRL click any or multiple servers to select them a button in the region where the server text is appears to delete all selected servers.

I kinda like this idea and would be down to make it if I can. I also like the drag reorder. 
Ill think about it and possibly make a pr with it soon. Idk yet though.

Leymooo:
right click popup menu for deleting and some more actions (like refresh/edit/add new server)?

disconnectedgrinders:
@Moulberry tryna figure out some designs for this, are you okay with having hover animations within the Pandora? Or is that outside of your design language?

Moulberry:
It depends on the animation but probably no

ItzELECTR0:
> are you okay with having hover animations within the Pandora? Or is that outside of your design language?

I know it's a bit late but I'd want to chip in that I think Pandora is one of the more well performant launchers so hover animations might eat into resources a bit more than necessary. I am personally not against it as long as there's options to disable them, but it's ultimately up to Moulberry. That aside, I really like the idea of being able to add/edit/remove servers directly from the launcher.

PR #459, opened 2026-04-29
fix: stop skin animation buttons from jittering

catpouch:
updates styling on the yaw/pitch/animation sliders in the skin preview so that the play/pause buttons don't jitter around when the yaw and pitch values are changing.
wanted to get involved in this project and i figured a small change was a good place to start!


Moulberry:
Thank you for your PR

PR #466, opened 2026-05-04
Translation hu

LitcherHUN:
Added hungarian locale, it should be 100% translated


Moulberry:
Thank you for your PR

PR #467, opened 2026-05-04
add sorting to modlist (name, size, data modified)

### LINKED ISSUE #456: [Feature Request] Mod sorting changes
Mods are currently sorted by the jar name, which sometimes is different from the actual mod name. For example, the jar of the mod "Create: The Factory Must Grow" is called tfmg-1.2.0.jar, which means that it does not appear next to other mods whose names start with "create" and it does not appear in search results for "create" or "factory". I would like if mods were sorted by their actual name and not by the jar's name.

Also I would like an option to sort enabled mods before disabled mods.

5sig2:
Added sorting options to the mod list, filename, displayname, size & date modified

closes https://github.com/Moulberry/PandoraLauncher/issues/456


crates/frontend/src/component/content_list.rs
```diff
@@ -436,14 +452,36 @@ impl ContentListDelegate {
...
         for modification in new_content.iter() {
-            mods.push(modification.clone());
+            let sort_key_source: &str = match self.sort_key {
+                InstanceContentSortKey::Filename => modification.filename.as_ref(),
+                InstanceContentSortKey::Name => modification.content_summary.name.as_deref().unwrap_or(modification.filename.as_ref()),
+                InstanceContentSortKey::ModifiedTime | InstanceContentSortKey::FileSize => modification.content_summary.name.as_deref().unwrap_or(modification.filename.as_ref()),
+            };
+            let sort_key: Arc<str> = sort_key_source.to_ascii_lowercase().into();
+
+            let (modified_unix_ms, file_size) = std::fs::metadata(modification.path.as_ref())
```
Moulberry, line 475: We want to avoid filesystem operations on the frontend. As it stands, this code will check the file metadata for every file every single time the "Mods" tab is opened. If there are hundreds of files in the directory, this could be bad

crates/frontend/src/component/content_list.rs
```diff
@@ -436,14 +452,36 @@ impl ContentListDelegate {
...
+        }
+
+        let mut items = Vec::with_capacity(new_content.len());
 
         for modification in new_content.iter() {
-            mods.push(modification.clone());
+            let sort_key_source: &str = match self.sort_key {
+                InstanceContentSortKey::Filename => modification.filename.as_ref(),
+                InstanceContentSortKey::Name => modification.content_summary.name.as_deref().unwrap_or(modification.filename.as_ref()),
+                InstanceContentSortKey::ModifiedTime | InstanceContentSortKey::FileSize => modification.content_summary.name.as_deref().unwrap_or(modification.filename.as_ref()),
```
Moulberry, line 471: These two branches seem the same? Might want to combine them if that is intentional

crates/frontend/src/component/content_list.rs
```diff
@@ -527,15 +560,70 @@ impl ContentListDelegate {
...
+        for item in &mut items {
+            item.children.sort_by(|a, b| {
+                if enabled_first && a.enabled != b.enabled {
+                    return b.enabled.cmp(&a.enabled);
+                }
+
+                let key_a: &str = match sort_key {
+                    InstanceContentSortKey::Filename => a.path.as_ref(),
+                    InstanceContentSortKey::Name => a.summary.name.as_deref().unwrap_or(a.path.as_ref()),
+                    InstanceContentSortKey::ModifiedTime | InstanceContentSortKey::FileSize => a.path.as_ref(),
```
Moulberry, line 606: This `time | size` case should probably be `a.summary.name.as_deref().unwrap_or(a.path.as_ref())` to match the behaviour of the sorting of regular mods

crates/frontend/src/component/content_list.rs
```diff
@@ -527,15 +560,70 @@ impl ContentListDelegate {
...
+                };
+                let key_b: &str = match sort_key {
+                    InstanceContentSortKey::Filename => b.path.as_ref(),
+                    InstanceContentSortKey::Name => b.summary.name.as_deref().unwrap_or(b.path.as_ref()),
+                    InstanceContentSortKey::ModifiedTime | InstanceContentSortKey::FileSize => b.path.as_ref(),
+                };
+
```
Moulberry, line 613: We can compare the FileSize for children if the ModpackFile's source is `ModpackFileSource::DownloadUrl` or `ModpackFileSource::Builtin`

crates/frontend/src/interface_config.rs
```diff
@@ -61,6 +65,27 @@ pub struct InterfaceConfig {
...
 
+#[derive(Debug, Clone, Copy, Default, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum InstanceContentSortKey {
+    #[default]
+    Filename,
+    Name,
+    ModifiedTime,
+    FileSize,
+}
```
Moulberry, line 76: Can you add "Mod ID" as a sort key, and then set the default to Name.
The fallback order should be Name -> Mod ID -> Filename.

So Filename should go:
Filename -> Name -> Mod ID

Name should go:
Name -> Mod ID -> Filename

Mod ID should go:
Mod ID -> Name -> Filename

crates/frontend/src/interface_config.rs
```diff
@@ -61,6 +65,27 @@ pub struct InterfaceConfig {
...
+
+impl InstanceContentSortKey {
+    pub fn name(self) -> SharedString {
+        match self {
+            InstanceContentSortKey::Filename => "File name".into(),
+            InstanceContentSortKey::Name => "Mod name".into(),
+            InstanceContentSortKey::ModifiedTime => "Modified time".into(),
+            InstanceContentSortKey::FileSize => "File size".into(),
```
Moulberry, line 84: "File name" -> Filename
"Mod name" -> Name
"Modified time" -> Modified Time
"File size" -> File Size

Also make translation keys for these

crates/frontend/src/interface_config.rs
```diff
@@ -79,6 +104,8 @@ impl Default for InterfaceConfig {
             page_path: Default::default(),
             quick_delete_mods: Default::default(),
             quick_delete_instance: Default::default(),
+            instance_mods_sort_key: Default::default(),
+            instance_mods_enabled_first: Default::default(),
```
Moulberry, line 108: Change this to instance_content_sort_key/instance_content_enabled_first.
I know this PR doesn't do it, but I'm gonna refactor the resourcepacks and mods page into a single "InstanceContentSubpage", and when I do that I'll add sorting to resourcepacks as well

crates/frontend/src/pages/instance/mods_subpage.rs
```diff
@@ -39,13 +41,32 @@ impl InstanceModsSubpage {
...
 
         let mods = instance.mods.clone();
 
+        let sort_dropdown = cx.new(|cx| {
+            let items = [
+                InstanceContentSortKey::Filename,
+                InstanceContentSortKey::Name,
+                InstanceContentSortKey::ModifiedTime,
+                InstanceContentSortKey::FileSize,
+            ].into_iter().map(|key| {
```
Moulberry, line 58: Use strum::IntoIter on the InstanceContentSortKey

Moulberry (changes requested):

Moulberry:
Merging, I'll make the changes myself since they're fairly simple. Thank you for your PR

PR #469, opened 2026-05-05
fix silent instance loading skip if file found in directory

catpouch:
fixes a bug where having any file in your instances directory could cause some instances to fail to load, since `BackendState::load_all_instances` would return early after hitting a file. i assume returning here was just a typo, but if there's some deeper reason i missed just ignore this.


Moulberry:
Thank you for your PR

PR #472, opened 2026-05-06
Flatpak package: Update release version to v5.1.0

meowIAmAUser:
Was still stuck on v2.5.0


Moulberry:
I don't even use this. I just accepted a PR that added it because the flatpak author asked. If the author of the flatpak wants to change this, they can submit a PR. If they need it updated every release they could ask for me to do that

PR #481, opened 2026-05-11
Add Swedish translations

enkvadrat:
I was not sure if I had to add translations where the term is the same in English, 
but I did so cause the others had done so.

After I made the translations I wanted to see how they looked in the actual interface, but it seems like there is 
no code for actually setting the language, I believe we can use: https://crates.io/crates/sys-locale to detect the current language one the system.

I added 
```rs
#[rustfmt::skip]
```
so that you can run rustfmt to format the repo without this generated code jumping around. I have noticed that a lot of source files are not formated using rustfmt (annoying if I want to modify code since my editor will format the code and the diff will contain a lot of unrelated changes) Is there any other reason that files in the repo is not formated?

### DIFF
--- a/crates/t/build.rs
+++ b/crates/t/build.rs
@@ -3,7 +3,7 @@ use std::{error::Error, path::Path};
 use indexmap::IndexMap;
 
 fn main() {
-    println!("cargo:rerun-if-changed=./locales.yml");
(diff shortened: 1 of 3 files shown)


Moulberry:
Thank you for your PR

PR #501, opened 2026-05-13
fix: update German translations

kehessen:
- change Modrinth translations to the official ones
- small typos fixed
- more natural wording
- add actual folder names to translated ones

(I feel like "Aktualisierung" could just be changed back to "Update" but don't know how others feel about that)


Shadowbee27:
I looked it up in the duden and update is there so I'd say update could be used instead of Aktualisierung

Moulberry:
Thank you for your PR

PR #522, opened 2026-05-24
fix: Forward XAUTHORITY on bwrap sandboxing

Awakened-Redstone:
The sandbox adds the path for XAUTHORITY (if set) but doesn't forward the variable, leading to failure to create the display if it is set.

I've been working on getting the sandbox to work on NixOS, and found that this was leading to failure on creating the display. Since it seems to be an issue that could affect more then just NixOS users I though I should make this a PR


Moulberry:
Thank you for your PR

PR #538, opened 2026-06-13
Fix AtLauncher override account not importing correctly.

dragmine149:
For some reason, atlauncher uses the the `username` field in place of the actual id *which i didn't fully catch the first time i did this in #299*

Hence this fixes that. However it does mean accounts have to be imported at the same time, just because we need to cross-reference the account to get the correct account.

For those who come across this PR, and have an issue where you need to always log-in to an instance, chances are the override account got incorrectly set and the only real way to fix that is to manually override it.


Moulberry:
Thank you for your PR

PR #557, opened 2026-06-28
Extract `format_downloads` to lib, reuse on pages

Simek:
# Why

Spotted that exact same `format_downloads` helper is defined in two pages code.

# How

Extract `format_downloads` to crate lib, reuse on pages.


Moulberry:
Thank you for your PR

PR #558, opened 2026-06-28
Slightly deemphasize mod lists entries tags

Simek:
# Why

It seems for me that there is not enough contrast between mod lists entries content, and everything seems to blend into continuous text wall.

# How

Slightly deemphasize mod lists entries tags by reducing font size and applying muted foreground color. As a part of the changes I have also tweaked spacing a bit, switched the Curseforge tags separator to bullet character, and used already locally defined `theme` where possible.

# Preview

Screenshot 2026-06-28 at 18 46 00
Screenshot 2026-06-28 at 18 46 10


Moulberry:
Thank you for your PR

PR #561, opened 2026-06-28
Optimize app icon SVG assets

Simek:
# Why

App icon SVG assets contains a lot of garbage data added by vector editors.

# How

Run app icon assets through SVGO (with safe config) to optimize SVG files. Make sure that there are no difference in icons appearance after optimization.


Moulberry:
Having the extra metadata makes them easier to edit. The difference is pretty tiny anyways, but if you wanted to optimize it it'd have to be temporarily at build time or something

PR #567, opened 2026-06-30
Rename shaderpack settings file on update

Artezon:
When shaderpack is updated through the launcher, its file name usually changes but `<shader_file>.txt` files are left orphan, now they will be renamed so users don't need to do this manually or configure shader again.


Moulberry:
Thank you for your PR

PR #568, opened 2026-07-01
Language switcher and system locale detection

### LINKED ISSUE #482: Detect system locale (langauge switching)
I created Swedish translations #481  but there was no way to actually use them.

The best way in my opinion is to detect the system locale using a crate like <https://crates.io/crates/sys-locale>

I guess we could also have a override for language switcher like in PrismLauncher.
However a lot of apps do not even have this so I do not think it is required, the best part about PrismLaunchers page is it links to a website where users can contribute translations themselves.

Image

Artezon:
Adds a language dropdown menu in settings (default - system language). Uses `sys-locale` crate for detection.

Closes #482

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7665,6 +7665,7 @@ name = "t"
 version = "5.2.2"
 dependencies = [
  "indexmap 2.13.0",
+ "serde",
  "sys-locale",
  "toml 0.9.12+spec-1.1.0",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7665,6 +7665,7 @@ name = "t"
 version = "5.2.2"
 dependencies = [
  "indexmap 2.13.0",
+ "sys-locale",
  "toml 0.9.12+spec-1.1.0",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,6 +96,7 @@ memchr = "2.7.6"
 toml = "0.9.11"
 clap = { version = "4.5.54", features = ["derive"] }
 open = "5.2.0"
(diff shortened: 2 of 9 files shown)


Artezon:
I want to contribute a translation as well as add previously hardcoded strings to translation file. Should it go to this PR or a new one?

enkvadrat:
> I want to contribute a translation as well as add previously hardcoded strings to translation file. Should it go to this PR or a new one?

You would usually do that in a separate PR.

enkvadrat:
Hm, I think that instead of having a string for "system" that behaves a lot different than the other ones, that could be an enum variant, like  
```rs 
enum Language {
    System,
    Specific(String),
}
```

Artezon:
> Hm, I think that instead of having a string for "system" that behaves a lot different than the other ones, that could be an enum variant, like
> 
> ```rust
> enum Language {
>     System,
>     Specific(String),
> }
> ```

Thank you, I changed to enums and it should be better now.  The `"system"` strings are only used for json serializing:
```rust
#[derive(Clone, Debug, PartialEq, Default, serde::Serialize, serde::Deserialize)]
#[serde(from = "String", into = "String")]
pub enum Language {
	#[default]
	System,
	Code(String),
}

impl From<String> for Language {
	fn from(s: String) -> Self {
		match s.as_str() {
			"system" => Language::System,
			_ => Language::Code(s),
		}
	}
}

impl From<Language> for String {
	fn from(lang: Language) -> Self {
		match lang {
			Language::System => String::from("system"),
			Language::Code(code) => code,
		}
	}
}
```

Moulberry:
Thank you for your PR

PR #569, opened 2026-07-04
Use text ellipsis on mod description

Simek:
# How

Use text ellipsis on mod description, it does not always work correctly with current tile layout, but still it feels better than the overflow cutoff.

# Preview

### Before

Screenshot 2026-07-04 at 11 34 13

### After

Screenshot 2026-07-04 at 11 33 49


Moulberry:
Text ellipsis is very broken in gpui... maybe they've improved it but I just encountered too many issues with it, the overflow might look a bit more weird but at least the layout doesn't break

PR #571, opened 2026-07-05
Skins menu enhancements

Artezon:
- Can now add skins via drag-and-drop
  

- Can change sort order (currenty selected skin is always first)
  
  

- Fixed wrong card size of 64x32 skins in texture mode and redundant spacing between rows
  Before
  
  After
  

- Made player model centered, this is visible when window is small (before/after)


Moulberry:
Thank you for your PR

PR #574, opened 2026-07-07
Fix Forge library loading crashes (ENOTDIR, relative URL panic, false hash mismatches)

Max-pro273:
Fixes two crashes during Forge install & launch:
- LoadLibrariesError(IoError(NotADirectory))
- LoadLibrariesError(Reqwest(RelativeUrlWithoutBase))
Root causes, all in `do_libraries_load` / `create_forgelike_install_version_modern`:
1. maven/ zip extraction didn't skip directory entries, writing them as 0-byte files that block the real jar's parent dir (ENOTDIR).
2. `do_libraries_load` called reqwest on libraries with empty/relative urls (locally-bundled by Forge), crashing instead of skipping the download.
3. Those same no-url libraries were still sha1-checked against a hash for the published artifact, which a locally-built file can never match (Prism Launcher works around this exact issue with -Dforgewrapper.skipHashCheck=true.)

Tested on Fedora (Forge 1.20.1, Forge 1.21.1, Forge 26.2)

### DIFF
--- a/crates/backend/src/launch/mod.rs
+++ b/crates/backend/src/launch/mod.rs
@@ -592,6 +592,11 @@ impl Launcher {
         // Extract files in maven/ into libraries, used in 1.16 and below
         for entry in installer_zip.entries() {
             if let Some(path) = entry.name.strip_prefix("maven/") {
+                // dir entries (e.g. "maven/.../1.20.1-47.4.20/") would otherwise
+                // get write_safe'd as a 0-byte file, blocking the real jar's parent dir (ENOTDIR)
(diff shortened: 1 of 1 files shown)


Moulberry:
The PR includes a lot of useless comments your LLM added. The has_downloadable_url change is just replacing one confusing error with another confusing error, which shouldn't even be happening in the first place.

I feel bad but I made this change without the extra noise in this commit: 713d877aca429b2094840f91f9145be0c5bfd364

Thank you for the PR and investigation

PR #575, opened 2026-07-10
Add more export options

Artezon:
Folders like backups and screenshots can be very big and usually I don't want to include them in modpack. This PR adds a possibility to exclude them. Fabric cache is also excluded if cache is not selected. Shaders are now handled similar to mods and resourcepacks, and are hashed by modrinth/curseforge.


Moulberry:
Thank you for your PR

PR #577, opened 2026-07-10
Add Russian translation and localize some hardcoded strings

Artezon:
Added Russian locale support, fixed "System language" dropdown text not updating to a new language upon switching.

Migrated hardcoded English strings across the frontend crate to use the localization system. However, hardcoded messages from other crates like backend are untouched. IMO they should be handled in frontend for translation, but I'm not sure how to implement that.

I added `.search_placeholder(t::common::search())` to any search bar so the text `Search...` is localized, but if there's a better way, like setting it in a single place, pls let me know, I'd be happy to change it :)

Note: I've changed `[common.custom]` to `[common.custom_icon]`. The reason is that all singular nouns in Russian and some other languages are gendered, and related adjective should be agreed with noun's gender. `[common.custom]` was only used in one place where it's related to custom icon, but if it would ever be related to something else it might have a different form because of a different gender or plural noun.

Hm, maybe `Custom icon` could be changed to `From file...` or `Browse...` considering it opens a file picker?


Moulberry:
Thank you for your PR

PR #578, opened 2026-07-10
Fall back to English instead of panicking when unsupported language is detected

Rajdacz:
Closes #579 

I noticed that when I was trying to launch the launcher, my system language (Polish, which is not supported) caused the panic immediately

I changed it to fall back to English instead of panicking


Moulberry:
Already fixed in 94d9455679e963fa0a159aea81c57d280649d837

PR #581, opened 2026-07-11
ui: fix Sync page open folder button overflow

Simek:
# How

Fix Sync page open folder button overflow for languages with longer label.

# Preview

### Before

Screenshot 2026-07-11 at 17 13 13

### After

Screenshot 2026-07-11 at 17 11 27


Moulberry:
Is the extra div necessary or does switching to min_h_72 enough?

Simek:
Without extra div the button stretches to 100% width since parent container with flex column uses `size_full`.

Moulberry:
Fixed in 13212842d38169a28275a92fd159e232d159dec7 using items_start instead of an extra div

PR #585, opened 2026-07-13
Improve unique numbered filenames generation

Artezon:
This logic for adding skins was repeated 2 times:
```rust
if path.exists() {
    for i in 1..32 {
        let new_filename = format!("{filename} ({i})");
        let new_path = self.directories.skin_library_dir.join(&new_filename);
        if !new_path.exists() {
            path = new_path;
            break;
        }
    }
}
```

- I extracted the logic into `lib.rs` so it can be reused across the backend
- If a file has extension, number will be placed before it (`skin (1).png`, not `skin.png (1)`)
- Does not overwrite old file if all numbered names within an attempted range are taken, instead appends a uuid
- Increased number of attempts from 32 to 100 (don't think someone will ever hit that limit but why not)

### DIFF
--- a/crates/backend/src/backend_handler.rs
+++ b/crates/backend/src/backend_handler.rs
@@ -13,7 +13,7 @@ use ustr::Ustr;
 use uuid::Uuid;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


Moulberry:
Thank you for your PR

PR #587, opened 2026-07-13
feat: duplicate instance

### LINKED ISSUE #360: [Feature Request] Add copy instance button
Allow the user to copy an instance if i for example want to upgrade that instance but also keep the old one I copy it/duplicate it modify and the other one.

Artezon:
Closes #360

~~Uses `copy_content_recursive`. Made this function cancellable.~~

Edit: when implementing this initially I forgot about deduplication aka hard linking with content library lol. So I decided to copy the function to `duplicate.rs` and make changes there to support it. `copy_content_recursive` is untouched now.

||||
|---|---|---|


MBgaming1:
This would be so useful, especially to change an instance to a newer version and update all the mods so you don't have to do it manually every time

crates/backend/src/duplicate.rs
```diff
@@ -0,0 +1,271 @@
...
+        filename
+    };
+    let dot = base.rfind('.')?;
+    Some(&base[dot + 1..])
+}
+
+fn hash_file(path: &Path, check_cancel: &dyn Fn() -> std::io::Result<()>) -> std::io::Result<[u8; 20]> {
+    let mut file = std::fs::File::open(path)?;
+    let mut hasher = sha1::Sha1::default();
+    let mut buf = vec![0_u8; 128 * 1024];
```
Moulberry: Might be better to make the scratch buffer an argument that way it doesn't need to be reallocated

crates/backend/src/duplicate.rs
```diff
@@ -0,0 +1,271 @@
...
+        if read == 0 {
+            break;
+        }
+        hasher.update(&buf[..read]);
+    }
+    Ok(hasher.finalize().into())
+}
+
+fn copy_file(from: &Path, to: &Path, check_cancel: &dyn Fn() -> std::io::Result<()>) -> std::io::Result<u64> {
+    let mut buf = vec![0_u8; 128 * 1024];
```
Moulberry: Same here, argument for the scratch buffer

crates/backend/src/duplicate.rs
```diff
@@ -0,0 +1,271 @@
...
+    let mut src = std::fs::File::open(from)?;
+    let mut dst = std::fs::File::create(to)?;
```
Moulberry: This newly created file doesn't have the same permission bits/attributes as the original, which could break executable files on Linux for example.

Copying a file manually like this is also much less efficient on CoW filesystems, e.g. apple, which can otherwise do the copy instantly.

We could maybe use the https://crates.io/crates/reflink-copy crate to attempt a reflink copy first, and if that fails then fall back to a slow copy

(Unrelated to this PR but I should probably also try to use reflink-copy instead of hard linking if possible)

crates/backend/src/duplicate.rs
```diff
@@ -0,0 +1,271 @@
...
+    for directory in directories {
+        _ = std::fs::create_dir(to.join(directory));
+    }
+
+    let mut files_done = 0_u64;
+    for (relative, source_path, is_library_eligible) in &files {
+        check_cancel()?;
+        let dest = to.join(relative);
+
+        if *is_library_eligible {
```
Moulberry: Again if reflinking is supported we can skip hashing the file etc.

Also can check nlink (unix) / number_of_links (windows) on the file metadata, and if there is only 1 hard link (the file itself) then we can skip hashing

Technically this code is problematic for security reasons, since it allows hard linking mods to writeable folders, but that's a bit complicated so I can write the fix for that

Moulberry (changes requested):

crates/backend/src/duplicate.rs
```diff
@@ -0,0 +1,271 @@
...
+    progress: &dyn Fn(u64, u64),
+    check_cancel: &dyn Fn() -> std::io::Result<()>,
+) -> std::io::Result<()> {
+    let from = from.canonicalize()?;
+    if !from.is_dir() {
+        return Err(ErrorKind::NotADirectory.into());
+    }
+    if !to.is_dir() {
+        return Err(ErrorKind::AlreadyExists.into());
+    }
```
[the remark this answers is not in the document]
Moulberry: create_dir is always called before the copy, I guess maybe the create_dir could be moved inside here but I think it's fine

crates/frontend/src/modals/duplicate_instance.rs
```diff
@@ -0,0 +1,129 @@
...
+                    this.default_name.as_str()
+                } else {
+                    text.as_str()
+                };
+                if resolved.is_empty() || !is_valid_instance_name(resolved) {
+                    this.name_invalid = true;
+                    return;
+                }
+                this.name_invalid = instance_names.contains(&text);
+            })
```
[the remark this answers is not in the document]
Moulberry: This should be fine since default_name shouldn't be invalid

crates/frontend/src/modals/duplicate_instance.rs
```diff
@@ -0,0 +1,129 @@
...
+        };
+
+        Self {
+            instance_id,
+            backend_handle,
+            name_input_state,
+            name_invalid: false,
+            default_name,
+            _name_input_subscription,
+        }
```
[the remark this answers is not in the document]
Moulberry: It should basically always find a free suffix, if it doesn't then it has to be empty

crates/frontend/src/lib.rs
```diff
@@ -313,6 +313,19 @@ pub(crate) fn is_single_component_path(path: &str) -> bool {
...
+        return original_name.to_string();
+    }
+    for i in 1..100 {
+        let numbered = format!("{original_name} ({i})");
+        if !existing_names.iter().any(|n| *n == &numbered) {
+            return numbered;
+        }
+    }
+    String::new()
+}
```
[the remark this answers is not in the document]
Moulberry: This is fine, creating the instance with an empty name should fail

crates/backend/src/duplicate.rs
Artezon: Done in 2ea33bd0ca0beb1fc16b83f87693a2bbd659058e

crates/backend/src/duplicate.rs
Artezon: Done in 2ea33bd0ca0beb1fc16b83f87693a2bbd659058e

crates/backend/src/duplicate.rs
Artezon: Done ^^
392eade8aea01757aa3e589071df546ea5fd668c 6d519a354ec4a4ab1949f53bfd19f0127dad87a2 (don't have mac or linux machine to test at the moment)

crates/backend/src/duplicate.rs
Artezon: 9b0a121dc800fae2b7b2bb00196683b0469978b3

The security issue is probably out of scope of this PR since it also affects install/update path (?)
I don't know unfortunately how to make hard links write-safe, maybe set read only attribute? But it seems to be easily circumvented, idk

crates/backend/src/duplicate.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    let source = {
+        let mut state = backend.instance_state.write();
+        let Some(instance) = state.instances.get_mut(id) else {
+            modal_action.set_error_message("Unable to duplicate instance, unknown id".to_string().into());
+            modal_action.set_finished();
+            return;
+        };
+        instance.root_path.clone()
+    };
```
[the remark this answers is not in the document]
Artezon: yeah... fixed now
cce443f28ed822c4c1c217a3106c97f7c5d02645

Artezon:
All fixed :)

Also I noticed that when I delete an instance with a lot of files it just stays at the same screen as if nothing is happening, until all files are deleted. Nothing seem to block users from starting a instance that is being deleted. A non-dismissable modal should be displayed during deletion.

Moulberry:
Thank you for your PR

PR #591, opened 2026-07-15
Add Ollama crash log analysis

imsawiq:
Adds Ollama crash-log analysis to the Logs page.

- Configure the Ollama Cloud key, endpoint and model in Settings.
- Click "Analyze with AI" next to the mclo.gs button.
- The response uses the launcher's selected language and supports Markdown.
- The API key is kept in the system keyring; log requests use the existing proxy-aware HTTP client.

Checked with `cargo check -p frontend -p backend`.


BrokenDecoder:
why there is a need of ai in this ?

imsawiq:
> why there is a need of ai in this ?

And why not? If you gets a crash, you can instantly figure out the cause with a single button press.

cosrnic:
> > why there is a need of ai in this ?
> 
> And why not? If you gets a crash, you can instantly figure out the cause with a single button press.

Or you get fed a ton of lies and when eventually that gets given to the mod/modpack authors it becomes incredibly annoying to deal with. LLMs are stupid and shouldn't be randomly inserted into something like this because of their confidence in lying.

bunchy7ss:
this is a silly pr with no real need.

Moulberry:
Unfortunately this is not a feature we are interested in right now

Issue #590, opened 2026-07-14
[Bug] Long mod names overflow past the column width


Artezon:
In instance content list, when mod name is too long, it pushes the next column to the right, but only in this row.

Suggestion: make both columns equal width, use ellipsis on overflow.

<img width="1571" height="737" alt="Image" src="https://github.com/user-attachments/assets/c057ad98-3ee9-4837-b35b-36a289e55740" />

BrokenDecoder:
Fixed in My 2nd commit 

<img width="1172" height="950" alt="Image" src="https://github.com/user-attachments/assets/c3ad44ad-a741-46f7-98f5-21b7279b7894" /> 
the same issue happens with the right sided delete button

PR #592, opened 2026-07-16
feat: Added automated Java Downloader, mclo.gs Log Upload, and UI  Stuff

BrokenDecoder:
**Automated Java Downloader**
>  Integrated a one-click Java installer that grabs the latest runtimes straight from the Adoptium API so users don't have to download it manually anymore.

**Log Uploading**
> Added a quick "Upload Logs" button in the game output window. It automatically posts the console logs to mclo.gs and copies the link to your clipboard

Screenshot 2026-07-16 190030
Screenshot 2026-07-16 190228


BrokenDecoder:
Fixed Isuee #590 
Screenshot 2026-07-16 195042

crates/frontend/src/game_output/mod.rs
```diff
@@ -1081,12 +1155,17 @@ impl Render for GameOutputRoot {
                 state.scrolling = GameOutputScrolling::Top { offset: Pixels::ZERO };
                 cx.notify();
             })))
+            .child(Button::new("upload").label("Upload Logs").on_click(cx.listener(|root, _, window, cx| {
```
Artezon: Please use translation file when adding strings to the `frontend` crate. For example, text for uploading logs already exists there and can be reused. Otherwise add it to `locales.toml` with (at least) an English translation.
```suggestion
            .child(Button::new("upload").label(t::instance::logs::upload::label()).on_click(cx.listener(|root, _, window, cx| {
```

Artezon:
> Fixed Isuee #590

The left column is wider for some reason, especially when window is maximized. But honestly I tried to make them equal and can't figure it out myself, perhaps this is a problem with the GPUI library 🤔

crates/frontend/src/game_output/mod.rs
Artezon: There are also other strings you added that need to be translatable

Moulberry:
Your mojang.svg is an entire webpage
How did you even manage to do that

PR #598, opened 2026-07-22
Feat: Add optional in-app Game Output tab for instances

dennisthegamer:
Route Minecraft game output into an in-app "Game Output" tab on the instance page instead of a separate window, controlled by a global setting (frontend InterfaceConfig) with a per-instance tri-state override (Default / Always in tab / Always in window). The backend routes window vs. tab at launch for all launch paths; the frontend holds each instance's output and shows the tab live based on the effective setting. Reuses the existing game-output renderer and adds en/de/hu/ru/sv strings.

image
image

### DIFF
--- a/crates/t/locales.toml
+++ b/crates/t/locales.toml
@@ -2737,3 +2737,45 @@ ru = "м"
 [time.s]
 en = "s"
 ru = "с"
+
+[settings.windows.terminal_in_tab]
+en = "Open game output in tab instead"
+de = "Spielausgabe stattdessen im Tab öffnen"
+hu = "Játéknapló megnyitása lapon helyette"
+ru = "Открывать логи игры во вкладке"
+sv = "Öppna spelutdata i flik istället"
+
+[instance.terminal.title]
+en = "Game Output"
+de = "Spielausgabe"
+hu = "Játéknapló"
+ru = "Логи игры"
+sv = "Spelutdata"
+
+[instance.terminal.no_output]
+en = "No output yet"
+de = "Noch keine Ausgabe"
+hu = "Még nincs kimenet"
+ru = "Пока нет вывода"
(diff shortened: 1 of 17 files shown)


5sig2:
THIS I LOVE THIS

Moulberry:
The change is fine but the code is a mess due to the usage of AI

dennisthegamer:
> The change is fine but the code is a mess due to the usage of AI

Yeah, it's true. I've used AI. I thought I could do something good with it, but unfortunately it was the opposite.

BrokenDecoder:
> The change is fine but the code is a mess due to the usage of AI

lemme fix it

crates/t/locales.toml
```diff
@@ -2737,3 +2737,45 @@ ru = "м"
 [time.s]
 en = "s"
 ru = "с"
+
+[settings.windows.terminal_in_tab]
+en = "Open game output in tab instead"
+de = "Spielausgabe stattdessen im Tab öffnen"
+hu = "Játéknapló megnyitása lapon helyette"
```
LitcherHUN, line 2744: This line should be: "Új ablakban nyíljon meg a játéknapló"

crates/t/locales.toml
```diff
@@ -2737,3 +2737,45 @@ ru = "м"
...
+en = "Default (global)"
+de = "Standard (global)"
+hu = "Alapértelmezett (globális)"
+ru = "По умолчанию (глобально)"
+sv = "Standard (global)"
+
+[instance.terminal.mode.tab]
+en = "Always in tab"
+de = "Immer im Tab"
+hu = "Mindig lapon"
```
LitcherHUN, line 2772: This should be "Mindíg a fülben"

Moulberry:
Added this myself, thank you for the PR anyways

Issue #287, opened 2026-03-16
[BUG] Cant drag launcher from pandora logo


disconnectedgrinders:
Be able to drag the launcher window while clicking and holding down on the pandora logo as this works everywhere except the sidebar.
Kinda unsure if this is a bug or not, but I assume it is.
OS: Windows 11 25H2
Version: 4.0.1

PR #612, opened 2026-08-05
Make Pandora logo area draggable like title bar

### LINKED ISSUE #610: [Enhancement] Make Pandora logo area draggable like title bar
We can already use title bar to move the window and maximize on double click, but if window is small, would also be convenient to extend it to this area:

Image

Artezon:
Closes #610

Tested on windows
Needs proper testing on x11, wayland, macOS

### DIFF
--- a/crates/frontend/src/component/title_bar.rs
+++ b/crates/frontend/src/component/title_bar.rs
@@ -15,8 +15,8 @@ pub struct TitleBar {
 }
 
 #[derive(Default)]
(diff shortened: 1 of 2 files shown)


henbotb:
Not sure why, but Moulberry had previously closed an issue for this as not planned.
https://github.com/Moulberry/PandoraLauncher/issues/287

Moulberry:
Thank you for your PR

PR #614, opened 2026-08-05
feat(translation): Addition of the French translation

julienallexandre:
Addition of the French translation 🇫🇷


Moulberry:
Not to be rude, but did you write this translation? There are some weird formatting quirks which makes it seem like it could be machine translation. For example replacing " - " with " : " and adding "« »" around mod names which isn't present in all the other translations. If this was written by you, why all the changes?

Another once is changing "Shift+Click" to "Appuyez sur Maj + clic" instead of "Maj+Clic"

julienallexandre:
I used DeepL for the bulk of it, but the translations were poor, so I made quite a few corrections. As for the formatting, I’ll sort that out. :)

julienallexandre:
I’ve removed the « » from the mod names.
I’ve corrected quite a few instances of ':' to '-'.
I’ve fixed the Shift+Click so that the prompt to press the key no longer appears.
I’ve also removed the ':' signs and replaced them with '-', which makes more sense in French. But I think this should be done for all languages. For example: 

_[settings.proxy.launcher_only_note]
en = "Note - Proxy settings only apply to the launcher, not the game itself"
de = "Achtung - Proxyeinstellungen sind nur für Pandora, nicht für Minecraft"
hu = "Megjegyzés - A proxy beállítások csak az indítóra hatnak, nem a játékra"
sv = "Observera - Proxy inställningar gäller endast för launchern inte för själva spelet"
**ru = "Примечание: настройки прокси применяются только к лаунчеру, а не к самой игре"**
fr = "Remarque : les paramètres de proxy s'appliquent uniquement au lanceur, et non au jeu lui-même."
fr = "Remarque - les paramètres de proxy s'appliquent uniquement au lanceur, et non au jeu lui-même."_

And just a quick aside on the multiple spaces between text and the ':' symbol, for example. This is standard in French.

Moulberry:
Thank you for your PR

PR #623, opened 2026-08-11
feat: add downloads (Server Resourcepack) folder sync

Maganoos:
Needs some translations, I only did the ones that I could personally do (even if they might be shit)


Moulberry:
Thank you for your PR

PR #624, opened 2026-08-13
Add an optional local control API over a Unix socket

uncaney:
## What this is

An optional, local control API for the launcher. It adds a second socket next to `launcher.sock` so the launcher can be driven programmatically: an NDJSON request/response protocol on a Unix domain socket at `<launcher_dir>/api.sock` (mode `0600`, same user only), served by the backend so it reads instance/account state directly and sends `MessageToBackend` like a second frontend.

The motivation: I wanted to script the launcher (spin up instances, install content, launch/kill, watch game output) from other tools without embedding the whole thing as a library. `launcher.sock` only forwards argv and focuses the window on every connection, so it is not usable for this. The new listener is silent and never touches the UI unless a command explicitly asks it to (`launcher.focus`).

## Scope

It wraps essentially the whole `MessageToBackend` surface:

- **Instances**: list/get/create/delete/rename/start/kill, every configuration field (version, loader, memory, JVM flags/binary, wrapper, Linux wrapper, system libraries, sandbox, icon), worlds/servers/content, export (zip/mrpack/curseforge), import (`.mrpack` and other launchers), relocate, shortcut.
- **Content**: search and version listing on Modrinth and CurseForge (so a client can discover before installing), install by id, enable/disable/delete/update, modpack children.
- **Accounts**: offline add, select, delete, reorder, Microsoft device-code login and re-auth, skins and capes, plus the skin library.
- **Sync, settings** (the proxy password is never returned), **logs** (list/read/upload/cleanup), **metadata**, **launcher self-update**.
- A generic async-operation model (`ops.list/status/cancel`) backed by `ModalAction`, so progress, errors and the login verification URL are pollable. A `game.output` event streams stdout/stderr per instance while a game runs.

The binary doubles as a client: `pandora_launcher --api <cmd> [--api-params <json>]` talks to the running launcher and prints the JSON response; `events.subscribe` streams. Full reference in `docs/CONTROL_API.md`.

## Implementation notes / changes outside the new file

Most of the code is the new `crates/backend/src/api_server.rs`. The rest is small hooks:

- `FrontendHandle` gains an optional message tap (shared across clones) so the API can re-emit state changes as events without disturbing the single-consumer frontend channel.
- Instance launch and Microsoft login are moved onto their own tasks. Previously they were awaited inline on the single backend message loop, so a slow launch (asset/Java downloads) or a login waiting on a browser redirect parked every other message until it finished. This is a good change on its own; it is also what keeps the API responsive during a launch.
- `DeleteInstance` now deregisters the instance directly after removing its folder instead of relying on the filesystem watcher. On macOS the watcher's removal event is dropped intermittently (FSEvents), which left a ghost instance in the registry until restart; the watcher path stays as a no-op backstop.
- `game.output` needs the live console stream, so `start_game_output` grew an optional tee (the frontend window still works unchanged when no client is subscribed), and `GameOutputMsg` is now `Clone`.

## Footprint and safety

- Unix only for now: the listener and the `--api` client are compiled out on Windows (the socket handling would need the named-pipe path). Everything else is untouched there.
- One new dependency, `md-5` (already vendored), for offline-account UUIDs, plus the tokio `io-util`/`time` features. `Cargo.lock` changes by two lines; `vendor/` is untouched.
- Threat model is a same-user local process; the socket is `0600` with no additional auth, which matches how `launcher.sock` already trusts the local user.

## Status

I have been running this on my own build for a while (macOS). It builds with the existing vendored `--frozen --offline` path. I am opening it as a proposal. If a control surface like this is something you would consider upstreaming, I am happy to iterate on the shape, the command naming, gating it behind a feature flag, or adding the Windows named-pipe path.


Moulberry:
This isn't a feature we're interested in right now and reviewing 3.5k lines of AI-generated code is going to take too long for me

PR #635, opened 2026-08-20
Fix modrinth db migration

mcbabo:
Since Modrinth changed their app.db schema, this is a fix for this migration:
https://github.com/modrinth/code/blob/3e07267e10ee687c25ccc5dfe3a77fdd00088000/packages/app-lib/migrations/20260611120000_instances-content-foundation.sql

This should consider both new `instances` and old `profiles` table setup.

You can still closes this pr if you want to make it yourself, just wanted to fix it for now :)

The migration includes:
```sql
INSERT INTO instances (
    id,
    path,
    applied_content_set_id,
    install_stage,
    launcher_feature_version,
    update_channel,
    name,
    icon_path,
    created,
    modified,
    last_played,
    submitted_time_played,
    recent_time_played
)
SELECT
    'legacy:' || path,
    path,
    'legacy:' || path || ':default',
    install_stage,
    launcher_feature_version,
    preferred_update_channel,
    name,
    icon_path,
    created,
    modified,
    last_played,
    submitted_time_played,
    recent_time_played
FROM profiles;

INSERT INTO instance_content_sets (
    id,
    instance_id,
    name,
    source_kind,
    status,
    game_version,
    protocol_version,
    loader,
    loader_version,
    created,
    modified
)
SELECT
    'legacy:' || path || ':default',
    'legacy:' || path,
    'Default',
    CASE
        WHEN linked_project_id IS NOT NULL
            AND linked_version_id IS NOT NULL
            AND linked_version_id != ''
            THEN 'modrinth_modpack'
        WHEN linked_project_id IS NOT NULL
            AND (linked_version_id IS NULL OR linked_version_id = '')
            THEN 'server_project'
        ELSE 'local'
    END,
    'available',
    game_version,
    protocol_version,
    mod_loader,
    mod_loader_version,
    created,
    modified
FROM profiles;
```

### DIFF
--- a/crates/backend/src/launcher_import/modrinth.rs
+++ b/crates/backend/src/launcher_import/modrinth.rs
@@ -14,6 +14,15 @@ struct ModrinthInstanceToImport {
     minecraft_folder: Arc<Path>,
 }
 
+fn table_exists(conn: &rusqlite::Connection, table_name: &str) -> bool {
+    conn.query_row(
+        "SELECT 1 FROM sqlite_master WHERE type='table' AND name = ?1",
+        [table_name],
+        |_| Ok(()),
+    )
+    .is_ok()
+}
+
 pub fn import_instances_from_modrinth(backend: &BackendState, import_job: ImportFromOtherLauncherJob, modal_action: &ModalAction) -> rusqlite::Result<()> {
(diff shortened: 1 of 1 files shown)


Moulberry:
Thank you for your PR

PR #637, opened 2026-08-20
Update channel selection and change version dialog

Artezon:
Added content update channel selection (release, beta, alpha). Supports both Modrinth and Curseforge.

Also the last commit fixes dropdown names staying the same on opened page when language is changed by adding a special kind of dropdown name that accepts a translation function.

Screenshot from instance settings page:
Screenshot

Edit: Mod changelog, upgrade/downgrade, shift-click for quick update
Image

Image

Image

Edit 2: Added version fallback


Moulberry:
Hmm I'm not sure if "Update Channel" should be a per-instance setting. Every mod uses it differently.
I think it might be better to have a popup when you check for updates saying "X number of mods have updates," and then have a section where it says "X additional mods have an Alpha update" with an appropriate warning to go along with it.

I will have to think about it some more though

Artezon:
### Known issues

1) There's a crash when opening changelog for this shader because of weird formatting https://modrinth.com/shader/2ks/version/0.0.5_2
So I found out the problem is within gpui-component, updating it to commit https://github.com/longbridge/gpui-component/commit/26cc9366abb27ccedce386ac99a615a8fa7018da fixes the crash, but I don't know how to properly use vendoring and got too many changed files, so I won't commit this

2) Curseforge [docs](https://docs.curseforge.com/rest-api/#accessing-the-service) say there is 50 results per page limit, but I set 5000 and it works. Still it might stop working any moment so need help addressing that. Maybe replace the dropdown with a scroll container on the left side of the dialog that can have infinite scrolling

3) > Hmm I'm not sure if "Update Channel" should be a per-instance setting. Every mod uses it differently.
    > I think it might be better to have a popup when you check for updates saying "X number of mods have updates," and then have a section where it says "X additional mods have an Alpha update" with an appropriate warning to go along with it.
    
    Hm good idea though I don't quite understand how it should work. What about beta versions
    May keep per-instance channel as the default one, but add per-mod overrides... for me personally stable versions are always fine for every mod, and I already implemented fallback for mods that don't yet have stable release

FireDasher:
Maybe it could be like there are multiple download buttons for mods with beta/alpha version, like release is green, beta is yellow, and alpha is orange. And there would be buttons for "Update all to latest stable", "Update all to latest beta" which updates mods with beta updates and mods with stable updates, and "Update all to latest alpha" which updates mods with stable updates, n mods with beta updates, and mods with alpha update. I'm not sure how this would work with the modrinth api though.

FireDasher:
The channel dropdown probably makes more sense since it is closer to how the api works.

Artezon:
> I'm not sure how this would work with the modrinth api though

@FireDasher, all that would work just fine, it's more of an UX problem... I tried to include your ideas while keeping everything as simple as possible for the users. So I suggest this design, hope you like it :3

Instead of an "Update channel" per-instance setting, have a per-instance or even global checkbox: "Suggest pre-release updates".

If disabled:
- <kbd>Update all mods (Number)</kbd> button only updates to stable
- For mod rows, the green download button doesn't replace the <kbd>Change version</kbd> button. Instead, it appears to the left, and goes away when clicked. So no need to press <kbd>Shift</kbd>

If enabled:
- <kbd>Update all mods</kbd> becomes a dropdown with 3 options:
  - <kbd>Release versions only (Number)</kbd>
  - <kbd>Release and beta versions (Number)</kbd>
  - <kbd>Release, beta and alpha versions (Number)</kbd>
- For each mod, a maximum of 3 new buttons can temporarily appear at the same time, from left to right:
  - Orange <kbd>Download alpha</kbd> if an alpha update is available
  - Yellow <kbd>Download beta</kbd> if a beta update is available
  - Green <kbd>Download release</kbd> if a release update is available

Did not want to break the minimalistic look of the launcher and turn it into a plane dashboard, haha
But since most people won't enable beta/alpha for their instances, it's alright, I guess

A technical detail - if "Suggest pre-release updates" is enabled, an update check sends 3 requests instead of one (negligible IMO)

Things to consider
- When clicking <kbd>Update all</kbd> including alpha or beta, show a warning dialog saying it's not recommended and can break compatibility or your worlds. Include a "Don't show again" checkbox
- Points 1 and 2 of the known issues comment

FireDasher:
That sounds good, although it might be simpler if it just only suggested releases and you can choose the alpha version on the mod version changer, although then you wouldn't know if it has an alpha. There could also be a channel option in the "add from Modrinth" screen, or maybe when you install the mod it shows a version selector and you can shift click to skip it or there's a seperate button that skips it

Artezon:
> when you install the mod it shows a version selector

It already does if "Install Latest" is unchecked, though it can't install unsupported versions AFAIK



> There could also be a channel option in the "add from Modrinth" screen

When "Install Latest" is enabled, it installs whatever is the latest one, including alpha, so this is valid concern, channel selection should be added

> might be simpler if it just only suggested releases and you can choose the alpha version on the mod version changer

This is the default behavior (when the "Suggest pre-release updates" is unchecked)

We could remove "Suggest pre-release updates" altogether and let users install them only through the changer, but then we would lose the "Update all to latest alpha" thing :(

FireDasher:
> but then we would lose the "Update all to latest alpha" thing :(

Yeah, I think the update all to latest alpha thing would be helpful for people who like having the latest version even if it breaks compatibility or has bugs

crates/backend/src/metadata/items.rs
```diff
@@ -346,10 +347,35 @@ impl MetadataItem for ModrinthVersionMetadataItem {
...
 }
 
+#[derive(Debug)]
+pub struct ModrinthChangelogMetadataItem<'a>(pub &'a ModrinthChangelogRequest);
+
+impl<'a> MetadataItem for ModrinthChangelogMetadataItem<'a> {
+    type T = ModrinthChangelogResult;
+
+    fn request(&self, client: &reqwest::Client) -> RequestBuilder {
+        let url = format!("https://api.modrinth.com/v2/version/{}", self.0.version_id);
```
Moulberry, line 357: Isn't this the same request as ModrinthVersionMetadataItem?

crates/backend/src/metadata/items.rs
```diff
@@ -645,6 +672,11 @@ impl<'a> MetadataItem for CurseforgeGetModFilesMetadataItem<'a> {
         if let Some(mod_loader_type) = self.0.mod_loader_type {
             req = req.query(&[("modLoaderType", mod_loader_type)]);
         }
+        if let Some(release_types) = &self.0.release_types {
+            for release_type in *release_types {
+                req = req.query(&[("releaseTypes", release_type)]);
```
Moulberry, line 677: Where is this from? I don't see it in the CurseForge docs

crates/backend/src/metadata/items.rs
Moulberry: Isn't this the same endpoint as ModrinthVersionMetadataItem?

crates/bridge/src/instance.rs
```diff
@@ -269,6 +269,26 @@ impl ContentType {
...
     }
+
+    pub fn modrinth_loaders(&self) -> Option<Arc<[ModrinthLoader]>> {
+        match self {
+            ContentType::Fabric => Some(vec![ModrinthLoader::Fabric].into()),
+            ContentType::Forge | ContentType::LegacyForge => Some(vec![ModrinthLoader::Forge].into()),
+            ContentType::NeoForge => Some(vec![ModrinthLoader::NeoForge].into()),
+            ContentType::ResourcePack => Some(vec![ModrinthLoader::Minecraft].into()),
+            ContentType::ShaderPack => Some(vec![ModrinthLoader::Iris, ModrinthLoader::Optifine, ModrinthLoader::Canvas].into()),
+            _ => None,
```
Moulberry: Could you explicitly name the contenttype instead of using _ ? That way if we add more content types in the future, we get a compile error here so we can ensure it correct.

Maybe also take in the instance's loader as a parameter, and make the return type non-optional.

Also you don't need `vec![loader].into()`, just `[loader].into()` works and potentially avoids an extra allocation

crates/frontend/src/entity/metadata.rs
```diff
@@ -71,15 +71,23 @@ impl FrontendMetadata {
...
-                if let FrontendMetadataState::Loaded { keep_alive, .. } = existing.read(cx) {
-                    if !keep_alive.as_ref().map(|k| k.is_alive()).unwrap_or(true) {
-                        this.backend_handle.send(MessageToBackend::RequestMetadata {
-                            request: request.clone(),
-                            force_reload: false,
-                        });
-                    }
+                let mut failed = false;
+                match existing.read(cx) {
+                    FrontendMetadataState::Loaded { result: Err(_), .. } => failed = true,
```
Moulberry, line 76: I don't like this change. I don't want the launcher to spam requests and ddos modrinth or fabric or minecraft just because an API is returning an error.

I think this code should be kept as-is, but maybe reduce the keep_alive's duration to a short period like 10s if an error occurred

crates/frontend/src/modals/change_version.rs
```diff
@@ -0,0 +1,758 @@
...
+        }),
+        _ => unreachable!(),
+    }
+}
+
+fn modrinth_version_item(version: &ModrinthProjectVersion) -> Option<VersionItem> {
+    if version.files.is_empty() {
+        return None;
+    }
+    if let Some(status) = version.status && !matches!(status, ModrinthVersionStatus::Listed | ModrinthVersionStatus::Archived) {
```
Moulberry: I'm pretty sure if the version is unlisted the API won't return it to us, so I think this check here can be removed

crates/frontend/src/modals/change_version.rs
```diff
@@ -0,0 +1,758 @@
...
+    }
+}
+
+fn display_game_version(game_versions: &[Ustr]) -> Option<String> {
+    let first = game_versions.first()?.as_str();
+
+    Some(match game_versions.len() {
+        1 => first.to_owned(),
+        2 | 3 => game_versions.iter().map(Ustr::as_str).join(", "),
+        _ => format!("{first}+"),
```
Moulberry: 1.21, 1.21.1, 121.2, 1.21.3 for example should probably show as 1.21-1.21.3 instead of 1.21+
1.21+ makes it look like it supports all the versions up to the latest

crates/frontend/src/modals/change_version.rs
```diff
@@ -0,0 +1,758 @@
...
+        let source = match project_id {
+            Some(project_id) => ContentSource::ModrinthProject { project_id },
+            None => {
+                self.items = Some(Vec::new());
+                return;
+            },
+        };
+        let mut hash = [0_u8; 20];
+        if hex::decode_to_slice(self.installed_sha1.as_ref(), &mut hash).is_ok() {
+            self.data.backend_handle.send(MessageToBackend::SetContentSource { hash, source: source.clone() });
```
Moulberry: So, this might need some context, the ModrinthUnknown content source isn't set anymore, it's just legacy from back when I didn't track the project id.

I don't think all this code on the frontend to discover the content source and then send the content source back to the backend is necessary, especially considering it is just legacy.

If you want, you can add some code on the backend to try to lookup the project id to get rid of the legacy entries

crates/backend/src/backend_handler.rs
```diff
@@ -631,50 +615,45 @@ impl BackendState {
...
+                                                            game_versions: [version].into(),
+                                                            version_types,
+                                                        }
+                                                    }).await
+                                                },
+                                            };
+
+                                            if !matches!(fetch_result, Err(MetaLoadError::NonOK(404))) {
+                                                return fetch_result;
+                                            }
```
Moulberry, line 644: I don't think this whole update logic works as intended.

Say our update channel is Release, then we're gonna get [[Release],[Beta],[Alpha]]
Lets say we have a mod with version 1.0 which is Release, but the mod has an update which is 1.1 and Beta.

So first it tries to find an update for Release, but because there is no update with Release, we get a 404. This code will then ignore the 404, and will then look for [Beta], which it will then find an update to.

So even though our update channel is Release and we're on the latest Release, we get updated to the Beta. 

If our update channel is Release, why would we be looking for Beta updates at all?
Except when our current jar is Beta, then it should be valid to update to another Beta even if our channel is Release. Some mods e.g. c2me only publish Alpha releases.

Moulberry (changes requested):

Moulberry:
Gonna request an AI review even though it's garbage it sometimes catches things

crates/frontend/src/modals/change_version.rs
```diff
@@ -0,0 +1,758 @@
...
+
+    fn discover_project(&mut self, cx: &mut Context<Self>) {
+        let request = FrontendMetadata::request(
+            &self.data.metadata,
+            MetadataRequest::ModrinthVersionsFromHashes(ModrinthVersionsFromHashesRequest {
+                hashes: vec![self.installed_sha1.clone()].into(),
+                algorithm: "sha1".into(),
+            }),
+            cx,
+        );
```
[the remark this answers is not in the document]
Moulberry: This whole method is getting removed

crates/frontend/src/modals/change_version.rs
```diff
@@ -0,0 +1,758 @@
...
+                .items_center()
+                .justify_center()
+                .text_color(cx.theme().muted_foreground)
+                .child(t::instance::content::change_version::no_changelog()).into_any_element(),
+            ChangelogDisplay::Loaded(Some(ChangelogContent::Markdown(changelog))) => TextView::markdown("changelog", changelog)
+                .scrollable(true)
+                .h_full()
+                .w_full()
+                .into_any_element(),
+            ChangelogDisplay::Loaded(Some(ChangelogContent::Html(changelog))) => TextView::html("changelog", changelog)
```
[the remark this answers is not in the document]
Moulberry: I had this concern too but it only renders simple HTML.. hopefully..

crates/backend/src/metadata/items.rs
Artezon: It's from here: https://github.com/aternosorg/php-curseforge-api/blob/5482843cebf323c61b0e45bf5ebb7cb367c7c1ab/openapi.yaml#L599-L606

I tested and it actually works, so official docs are incomplete ig. Sorry, I forgot to tell you that I used an undocumented API here (._. )

Checked it again right now

Only releases:
```
curl -X GET "https://api.curseforge.com/v1/mods/394468/files?modLoaderType=4&pageSize=1&releaseTypes=1" -H "x-api-key: REDACTED"
```
```
{"data":[{"id":8396480,"gameId":432,"modId":394468,"isAvailable":true,"displayName":"Sodium 0.9.1 for Fabric 26.1.2","fileName":"sodium-fabric-0.9.1+mc26.1.2.jar","releaseType":1,"fileStatus":4,"hashes":[{"value":"cdb5ab59dc05840c5fc762c3821570c6fa02a8dc","algo":1},{"value":"f173cec7478bd6ea8b39ab3fbc769a3b","algo":2}],"fileDate":"2026-07-08T21:04:00.503Z","fileLength":1889595,"downloadCount":83574,"fileSizeOnDisk":3505055,"downloadUrl":"https://edge.forgecdn.net/files/8396/480/sodium-fabric-0.9.1%2bmc26.1.2.jar","gameVersions":["Client","Fabric","26.1.2"],"sortableGameVersions":[{"gameVersionName":"Client","gameVersionPadded":"0","gameVersion":"","gameVersionReleaseDate":"2022-12-08T00:00:00Z","gameVersionTypeId":75208},{"gameVersionName":"Fabric","gameVersionPadded":"0","gameVersion":"","gameVersionReleaseDate":"2022-09-01T00:00:00Z","gameVersionTypeId":68441},{"gameVersionName":"26.1.2","gameVersionPadded":"0000000026.0000000001.0000000002","gameVersion":"26.1.2","gameVersionReleaseDate":"2026-04-09T11:39:54.243Z","gameVersionTypeId":83806}],"dependencies":[],"alternateFileId":0,"isServerPack":false,"fileFingerprint":4270737160,"modules":[{"name":"META-INF","fingerprint":800631806},{"name":"net","fingerprint":3519194131},{"name":"config-icon.png","fingerprint":4001690648},{"name":"fabric.mod.json","fingerprint":777049291},{"name":"sodium-common.mixins.json","fingerprint":69899163},{"name":"sodium-fabric.accesswidener","fingerprint":3869862196},{"name":"sodium-fabric.mixins.json","fingerprint":374894530},{"name":"sodium-frapi.mixins.json","fingerprint":422552508},{"name":"sodium-icon.png","fingerprint":1658709492},{"name":"assets","fingerprint":818355870},{"name":"LICENSE.md","fingerprint":2033162507}]}],"pagination":{"index":0,"pageSize":1,"resultCount":1,"totalCount":85}}
```
Releases and alpha:
```
curl -X GET "https://api.curseforge.com/v1/mods/394468/files?modLoaderType=4&pageSize=1&releaseTypes=1&releaseTypes=3" -H "x-api-key: REDACTED"
```
```
{"data":[{"id":8591753,"gameId":432,"modId":394468,"isAvailable":true,"displayName":"Sodium 0.9.2-alpha.4 for Fabric 26.1.2","fileName":"sodium-fabric-0.9.2-alpha.4+mc26.1.2.jar","releaseType":3,"fileStatus":4,"hashes":[{"value":"85e6b194a891f90c6b84c34adaac19a793cc8d2f","algo":1},{"value":"a572e850e94be2700f44688981419cac","algo":2}],"fileDate":"2026-08-07T00:47:15.96Z","fileLength":1936354,"downloadCount":0,"fileSizeOnDisk":3612084,"downloadUrl":"https://edge.forgecdn.net/files/8591/753/sodium-fabric-0.9.2-alpha.4%2bmc26.1.2.jar","gameVersions":["Client","Fabric","26.1.2"],"sortableGameVersions":[{"gameVersionName":"Client","gameVersionPadded":"0","gameVersion":"","gameVersionReleaseDate":"2022-12-08T00:00:00Z","gameVersionTypeId":75208},{"gameVersionName":"Fabric","gameVersionPadded":"0","gameVersion":"","gameVersionReleaseDate":"2022-09-01T00:00:00Z","gameVersionTypeId":68441},{"gameVersionName":"26.1.2","gameVersionPadded":"0000000026.0000000001.0000000002","gameVersion":"26.1.2","gameVersionReleaseDate":"2026-04-09T11:39:54.243Z","gameVersionTypeId":83806}],"dependencies":[],"alternateFileId":0,"isServerPack":false,"fileFingerprint":524056701,"modules":[{"name":"META-INF","fingerprint":4215337759},{"name":"net","fingerprint":3943899613},{"name":"config-icon.png","fingerprint":4001690648},{"name":"fabric.mod.json","fingerprint":271192500},{"name":"sodium-common.mixins.json","fingerprint":3455656890},{"name":"sodium-fabric.accesswidener","fingerprint":3869862196},{"name":"sodium-fabric.mixins.json","fingerprint":374894530},{"name":"sodium-frapi.mixins.json","fingerprint":422552508},{"name":"sodium-icon.png","fingerprint":1658709492},{"name":"assets","fingerprint":536290699},{"name":"LICENSE.md","fingerprint":2033162507}]}],"pagination":{"index":0,"pageSize":1,"resultCount":1,"totalCount":96}}
```

crates/frontend/src/entity/metadata.rs
Artezon: The problem I tried to fix is when bad internet (steps to reproduce):
1. Disconnect ethernet cable or wifi
2. Click switch version and wait for error, then close modal
3. Connect again
4. Click switch version again - it will not even try, give up instantly, until I restart the launcher or time passes
5. This also the same for some other pages fetching online information

I will revert this and try to work with keep_alive duration instead

crates/frontend/src/modals/change_version.rs
Artezon: done
should I also add count, like `1.21-1.21.3, 4 total`?

crates/frontend/src/modals/change_version.rs
Artezon: Alright, I removed this code. To not confuse users with "No versions" message for legacy entries, I came up with this error message:

crates/backend/src/metadata/items.rs
Artezon: Yes, request is the same but the result saves different things: ModrinthProjectVersion vs ModrinthChangelogResult

I could just add "changelog" field to ModrinthProjectVersion but ModrinthProjectVersion is used in a lot of places and I thought changelog would be cached even when not needed, so lots of changelogs would just sit in RAM for no reason (they may be kilobytes)... maybe I have a wrong understanding though

Also it matches how Curseforge has separate changelog request

crates/bridge/src/instance.rs
Artezon: Added explicit names and removed vec!

I made modrinth_loaders non-optional, but kept curseforge_loader return value as-is because it would break updates for resourcepacks and shaders

crates/frontend/src/modals/change_version.rs
Artezon: Done. Hmm, you may want to check if this is needed then:
https://github.com/Moulberry/PandoraLauncher/blob/3f2cbba49faf4b7de427dc639e99e2e4f223d6b8/crates/frontend/src/modals/modrinth_install.rs#L121-L125

crates/backend/src/backend_handler.rs
```diff
@@ -717,25 +696,31 @@ impl BackendState {
...
-                                        ContentType::NeoForge => {
-                                            Some(CurseforgeModLoaderType::NeoForge as u32)
-                                        },
-                                        _ => None
-                                    };
+                                    let mod_loader_type = summary.content_summary.extra.curseforge_loader().map(|loader| loader as u32);
+
+                                    let result = async {
+                                        for &release_types in update_channel.curseforge_release_types_with_fallback().iter() {
+                                            let fetch_result = self.meta.fetch(&CurseforgeGetModFilesMetadataItem(
```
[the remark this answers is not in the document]
Artezon: well, it compiles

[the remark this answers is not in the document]
Artezon: I will change it to meta.fetch anyway

crates/frontend/src/modals/change_version.rs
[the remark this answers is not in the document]
Artezon: hmm pretty sure there's no JS engine whatsoever

or is there? *(vsauce music playing)*

crates/backend/src/metadata/items.rs
Moulberry: Hmm, perhaps. Although it does expire so I'd expect it to not sit in RAM forever. I generally like to minimize the number of network requests, since network requests are always the biggest failure point (especially when modrinth keeps going down). For now it's fine, I can revisit this later if needed.

crates/backend/src/backend_handler.rs
Moulberry: I just realized I was mistaken, if the version we have is already the latest the update API just returns the latest version again, not 404. So the code works. My mistake

crates/frontend/src/modals/change_version.rs
Moulberry: I think just the range is fine

Moulberry:
Still a couple things left to do, but they're simple enough that I can do them. Thank you very much for the PR

PR #640, opened 2026-08-23
add missing german translations and fix typo in quitting message

kehessen:
literally unusable with the current quit message


Moulberry:
Thank you for your PR

Issue #646, opened 2026-08-27
[Feature Request] CurseForge manual mod download assistant with 3rd-party limits


FuzjaJadrowa:
Some mods on Curseforge have downloads blocked for third party launchers. as a result, installing the modpack leads to incomplete file downloads or an error, forcing the user to search for each mod manually. I would like to see a popup to appear once the modpack download finishes, displaying a list of missing mods or resource packs along with their download links. i would also like a "Download All" button added to this popup that opens all the download links in the browser. launcher could then monitor the Downloads folder and import the files into the instance once it detects the correct ones.

Moulberry:
Added by #652

PR #652, opened 2026-08-29
Add manual CurseForge download support

NandeMD:
Some CurseForge files cannot be downloaded automatically due to restrictions. This PR adds a manual download workflow while keeping installation verification.

- Detects restricted CurseForge files during modpack installation.
- Shows a modal listing every required manual download.
- Supports opening individual pages or all download pages at once.
- Watches the selected downloads folder and validates files by hash.
- Moves verified files into the launcher cache and updates metadata.
- Adds manual download checking, folder selection, cancellation, and automatic completion.
- Handles duplicate files, stale/corrupt cache entries, and cross-filesystem moves.

And i think it solves #646


crates/backend/src/backend.rs
```diff
@@ -205,6 +206,213 @@ pub struct BackendState {
...
+            .unwrap_or(false)
+    };
+    if !temporary_valid {
+        if moved_source {
+            _ = tokio::fs::rename(&temporary, source).await;
+        } else {
+            _ = tokio::fs::remove_file(&temporary).await;
+        }
+        return Err(std::io::Error::other("Cached manual CurseForge download failed SHA-1 verification"));
+    }
```
Moulberry: So the code moves source -> temporary -> destination?
Why not just move source -> destination, I don't see what the extra temporary file is adding.

crates/backend/src/backend.rs
```diff
@@ -205,6 +206,213 @@ pub struct BackendState {
...
+
+    match tokio::fs::remove_file(&destination).await {
+        Ok(()) => {},
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {},
+        Err(err) => {
+            if moved_source {
+                _ = tokio::fs::rename(&temporary, source).await;
```
Moulberry: I prefer to use std::fs over tokio::fs. tokio::fs functions don't actually do the filesystem operation asynchronously, they just call std::fs inside tokio::task::spawn_blocking.

If the whole operation is expensive enough, then it's better to just make the whole function inside a tokio::task::spawn_blocking

crates/backend/src/backend.rs
```diff
@@ -229,6 +437,116 @@ impl CachedMinecraftProfile {
...
+
+        let result = session.wait().await;
+
+        if let Ok(paths) = &result {
+            for path in paths {
+                self.mod_metadata_manager.get_path(path);
+            }
+        }
+        self.manual_curseforge_downloads.lock().await.remove(&session_id);
+        _ = completion_send.send(());
```
Moulberry: completion_send should be a KeepAlive and completion should be a KeepAliveHandle

crates/backend/src/backend.rs
```diff
@@ -205,6 +206,213 @@ pub struct BackendState {
     pub quit_coordinator: QuitCoordinator,
     pub should_quit: AtomicBool,
     pub content_install_semaphore: Semaphore,
+    pub manual_curseforge_downloads: Mutex<HashMap<Uuid, Arc<ManualCurseforgeDownloadSession>>>,
+}
+
+pub struct ManualCurseforgeDownloadSession {
+    files: Arc<[ManualCurseforgeDownload]>,
+    completed_paths: Mutex<HashMap<[u8; 20], PathBuf>>,
```
Moulberry: I use parking_lot Mutexes over tokio mutexes for a couple reasons.
Firstly, they're faster.
Secondly, holding a mutex over an await point is usually bad and the non-send mutexes enforce this property

crates/backend/src/backend.rs
```diff
@@ -229,6 +437,116 @@ impl CachedMinecraftProfile {
...
+
+    pub async fn check_manual_curseforge_downloads(&self, session_id: Uuid) {
+        if let Some(session) = self.manual_curseforge_downloads.lock().await.get(&session_id) {
+            session.rescan.notify_one();
+        }
+    }
+
+    pub async fn cancel_manual_curseforge_downloads(&self, session_id: Uuid) {
+        if let Some(session) = self.manual_curseforge_downloads.lock().await.get(&session_id) {
+            session.cancelled.store(true, Ordering::Release);
```
Moulberry: Relying on the frontend to send cancel to trigger the cancel seems risky. Instead, we should send the frontend a KeepAlive and when this KeepAlive is dropped, it'll automatically set the cancelled flag to true

Moulberry:
Thank you for your PR

PR #653, opened 2026-08-30
Fix curseforge mods not updating when clicking update all

Artezon:
`can_update` was returning true only for modrinth


Moulberry:
Thanks I spotted this a few days ago but then forgot about it again lol

PR #654, opened 2026-08-30
Fix loading spinner staying on update error

Artezon:
When update fails, `updating.retain` will retain the loading state since filename hash does not change. Bug image:

Clears loading state when update either finishes or fails. Not sure, maybe there's a better way to fix this...


Moulberry:
Thank you for your PR

PR #656, opened 2026-08-30
Allow to scroll instance settings columns if height overflows

Artezon:


Moulberry:
Thank you for your PR

PR #663, opened 2026-09-06
fix(frontend): always open accounts sheet on click

theoneand33:
Removes the empty-accounts shortcut in the sidebar account button that skipped the sheet and started a Microsoft login directly.

The account selector sheet now always opens, even when not signed in. It already handles the empty state with Add / Add-offline buttons.


Moulberry:
This launcher doesn't support piracy
You have to log in with a microsoft account if you want to add offline accounts

PR #677, opened 2026-09-16
Translation update for Hungarian

LitcherHUN:
Added new lines

_(Btw at line 3680 you used Shift-Click instead of Shift+Click like everywhere else)_


Moulberry:
Thank you for your PR

Issue #678, opened 2026-09-17
[Bug] Sandboxed instances fail with Arch system Java


qualk:
## Description of bug
Sandboxed instances fail to launch when using Arch-packaged Java runtimes.

Arch packages JRE configs under a folder in `/etc/` according to their package name:

```text
/usr/lib/jvm/java-26-temurin/conf -> /etc/java-temurin
/usr/lib/jvm/java-25-temurin/conf -> /etc/java-25-temurin
/usr/lib/jvm/java-25-graalvm/conf -> /etc/java-25-graalvm
```

Pandora gives the sandbox read access to only the JRE directory. Java then cannot read `conf/security/java.security`:

```text
Exception in thread "main" java.lang.NoClassDefFoundError: Could not initialize class sun.security.jca.Providers
    at java.base/sun.security.util.SignatureFileVerifier.<init>(SignatureFileVerifier.java:128)
    at java.base/java.util.jar.JarVerifier.processEntry(JarVerifier.java:304)
    at java.base/java.util.jar.JarVerifier.update(JarVerifier.java:235)
    at java.base/java.util.jar.JarFile.initializeVerifier(JarFile.java:741)
    at com.moulberry.pandora.LaunchWrapper.main(LaunchWrapper.java:29)
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.InternalError: Error loading java.security file [in thread "main"]
    at java.base/java.security.Security$SecPropLoader.loadMaster(Security.java:143)
    at java.base/java.security.Security$SecPropLoader.loadAll(Security.java:121)
    at java.base/java.security.Security.initialize(Security.java:337)
    at java.base/java.security.Security.<clinit>(Security.java:326)
    at java.base/sun.security.jca.ProviderList.<init>(ProviderList.java:170)
    at java.base/sun.security.jca.ProviderList.fromSecurityProperties(ProviderList.java:89)
    at java.base/sun.security.jca.Providers.<clinit>(Providers.java:55)
```

## Steps to reproduce
1. Install a JDK/JRE whose configuration directory is an external symlink. On Arch, `jdk-temurin` provides `/usr/lib/jvm/java-26-temurin/conf -> /etc/java-temurin`.
2. Select `/usr/lib/jvm/java-26-temurin/bin/java` for an instance.
3. Enable instance sandboxing and launch the instance.

## This issue is unique
- [x] I've searched the other issues and didn't see an issue describing the same bug

## Environment
[...]

PR #679, opened 2026-09-17
Fix: #678

qualk:
Grant sandbox read access to the Java configuration directory when it resides outside the Java directory, fixing #678


Moulberry:
Thank you for your PR

PR #680, opened 2026-09-17
Translation update for Russian

Artezon:
Translated all new text to Russian, improved a few lines and added description for language field


Moulberry:
Thank you for your PR

PR #681, opened 2026-09-17
Added Missing Turkish Translations

NandeMD:
There were some missing turkish translations that will be added with this PR.


Moulberry:
Thank you for your PR

PR #683, opened 2026-09-19
Add global Java launch defaults

TreesOnTop:
Add global launch defaults for memory, JVM flags, and JVM binary. JVM flags are added to instance specific ones. The localization texts may need to be tweaked.


Moulberry:
Thank you for your PR

PR #686, opened 2026-09-20
Add setting to hide quickplay page

mukulx:
Added a toggle to hide the quickplay page under Settings > General.
image

image


Moulberry:
Thank you for your PR

